PR #7, opened 2018-03-21
Fix libsrtp2-dev installation on Ubuntu Xenial

it-svit:


jlaine:
I don't really want to go into advocating third party repos. I think a better solution would be to make pylibsrtp a compiled CFFI extension and ship a manylinux wheel for it. I've opened this ticket to track the issue:

https://github.com/jlaine/pylibsrtp/issues/2

moden-py:
Ok, any way thanks for the job.
Going to send PR with some new examples next week.

jlaine:
pylibsrtp now comes with prebuilt wheels, so you should not need to install libsrtp2-dev at all

PR #8, opened 2018-03-22
Add Docker to server example

it-svit:


jlaine:
Could you please explain why we would want Docker for an example? I'd like to keep the examples focused on aiortc itself, not all the different ways you can run a simple Python program.

jlaine:
Documenting the aiohttp dependency : I agree, I've updated the README.rst.

For the rest, it really seems like overkill to spin up a docker instance just to install two packages which are part of aiortc's base dependencies.

it-svit:
@jlaine, starting with docker shows that there is an easy way to run the application regardless of the operating system.
Thus, you can quickly see this example without installing dependencies in the main system.

jlaine:
Can you elabortate on what exactly we are making "easier"?

If it's installing the dependencies (libopus-dev and libsrtp-dev) we're not addressing the root cause of the problem, and a binary wheel which bundles these might be a better option.

For my information, what OS are you using that makes it so hard to install these dependencies?

jlaine:
Hi, I'm going to close this issue as I don't really see the added value in running an example in docker.

PR #9, opened 2018-03-22
Fix bug with undefined Video device

it-svit:
Hey, @jlaine.
I've found bug with situation when PC does not have any Video device.
I've fixed it by splitting query `navigator.mediaDevices.getUserMedia` on two queries.

### DIFF
--- a/examples/server/index.html
+++ b/examples/server/index.html
@@ -134,12 +134,32 @@ function start() {
         if (constraints.video) {
             document.getElementById('media').style.display = 'block';
         }
-        promise = navigator.mediaDevices.getUserMedia(constraints).then(function(stream) {
+
+        function addNewStream(stream) {
             stream.getTracks().forEach(function(track) {
                 pc.addTrack(track, stream);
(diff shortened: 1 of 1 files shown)


jlaine:
Could you explain a bit more why this is needed? If you don't have a video device, just don't check the "video" box and the call will be audio only, no?

These are examples, they need to be as readable as possible, so if something is not absolutely I'd rather not have it.

it-svit:
When User does not have any video device and will click on video checkbox, JS will raise exception. In this case, fetch request doesn't execute.

I've separated your logic to two certain queries and covered exception with `.catch()`.

it-svit:
Also, I've fixed bug with Stop button. (You wrote `getElementById` instead of `document.getElementById`)

jlaine:
The "stop" button now correctly appears / disappears as appropriate. I'm not in favour of silently ignoring errors, if you don't have a video device you can simply not tick the "video" box and you're done.

Issue #10, opened 2018-04-04
Video size mismatch


nickaknudson:
I'm working on sending and receiving video streams between python clients.

I am having an issue with a mismatch between the amount of bytes that I am sending and the amount that I am receiving. I was originally sending a 240h x 960w frame which is 345600 bytes, but on the receiving side I am getting 368640 bytes which I am then unable to marshal into the correct frame size.

I have simplified the example by sending a 6h x 8w frame which is only 72 bytes, but on the receiving side I am getting 864 bytes.

I think this might have something to do with the compression. I would appreciate some advice as to where to look.

Here is my working branch. I started by modifying cli.py https://github.com/nickaknudson/aiortc/blob/testing/examples/datachannel-cli/cli.py

nickaknudson:
It looks like `o_buf` is getting its size changed here:

https://github.com/jlaine/aiortc/blob/master/aiortc/codecs/vpx.py#L152

nickaknudson:
Here's some example debug output around line 152:
```
o_buf before line 152: len=72 o_pos=0 o_stride=8
o_buf after  line 152: len=160 o_pos=0 o_stride=8
o_buf before line 152: len=160 o_pos=8 o_stride=8
o_buf after  line 152: len=248 o_pos=8 o_stride=8
o_buf before line 152: len=248 o_pos=16 o_stride=8
o_buf after  line 152: len=336 o_pos=16 o_stride=8
o_buf before line 152: len=336 o_pos=24 o_stride=8
o_buf after  line 152: len=424 o_pos=24 o_stride=8
o_buf before line 152: len=424 o_pos=32 o_stride=8
o_buf after  line 152: len=512 o_pos=32 o_stride=8
o_buf before line 152: len=512 o_pos=40 o_stride=8
o_buf after  line 152: len=600 o_pos=40 o_stride=8
o_buf before line 152: len=600 o_pos=48 o_stride=4
o_buf after  line 152: len=644 o_pos=48 o_stride=4
o_buf before line 152: len=644 o_pos=52 o_stride=4
o_buf after  line 152: len=688 o_pos=52 o_stride=4
o_buf before line 152: len=688 o_pos=56 o_stride=4
o_buf after  line 152: len=732 o_pos=56 o_stride=4
o_buf before line 152: len=732 o_pos=60 o_stride=4
o_buf after  line 152: len=776 o_pos=60 o_stride=4
o_buf before line 152: len=776 o_pos=64 o_stride=4
[...]

PR #11, opened 2018-04-10
added videostream-cli example

nickaknudson:
This is a videostream-cli example that transmits VideoFrames between python clients.

### DIFF
--- a/aiortc/mediastreams.py
+++ b/aiortc/mediastreams.py
@@ -1,5 +1,7 @@
 import asyncio
+import cv2
 import math
+import numpy
 
 
 class AudioFrame:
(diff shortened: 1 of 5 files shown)


jlaine:
The more I look at this example, the more it feels as though it lacks a clear purpose. What are we actually demoing? Is it a three party video conference?

If we want to merge this I'd like:

- to completely roll back the changes in mediastreams.py
- to remove the unecessary workaround for #10  since it's fixed (yay!)
- to remove the VideoTrack_, all it does seems to be doable with 2 functions : Video frame -> BGR data and BGR data -> VideoFrame
- to do something a bit more exciting with the 3 "video tracks". At the moment we are doing some back and forth conversions between BGR and YUV (and dragging on opencv + numpy in the process) for little gain: we are just sending a fixed image.
- I have no problem writing received images to disk (it's actually a good thing, it's the only kind of output we have!), It's the SDP I don't really want written to disk

jlaine:
Hi Nick, I have commited a simplified version of the code in 3d7d471eb7dca90ecb872e0a38619829493ea2a4.

I still need to clean up the receiver's termination condition though.

PR #12, opened 2018-04-10
don't resize o_buf in VpXDecoder (fixes #10)

### LINKED ISSUE #10: Video size mismatch

nickaknudson:
I think this is correct. This prevents the `o_buf` array being resized. It appears that `i_stride` is usually bigger than `o_stride` so the slices were mismatched and `o_buf` was being resized.

### DIFF
--- a/aiortc/codecs/vpx.py
+++ b/aiortc/codecs/vpx.py
@@ -149,7 +149,7 @@ class VpxDecoder:
                     div = p and 2 or 1
                     o_stride = img.d_w // div
                     for r in range(0, img.d_h // div):
-                        o_buf[o_pos:o_pos + o_stride] = i_buf[i_pos:i_pos + i_stride]
+                        o_buf[o_pos:o_pos + o_stride] = i_buf[i_pos:i_pos + o_stride]
                         i_pos += i_stride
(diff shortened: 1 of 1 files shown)


jlaine:
That does looks right, and I'm actually puzzled by why the assignment even worked in the first place.

I ran your example, and the difference between i_stride and o_stride does seem to stem from a difference between vpx's *stored* image size and *display* image size:

img.w 1024 img.d_w 960
i_stride 1024 o_stride 960
i_stride 512 o_stride 480
i_stride 512 o_stride 480

PR #23, opened 2018-06-03
WIP - Add iceServers support at RTCConfiguration

### LINKED ISSUE #13: Pass STUN and TURN
I want to be able to pass STUN and TURN servers down to ICE Connection.

Happy to implement, would just like to understand your preference for the implementation. Should I pass to contructor of RPC Peer Connection or as some sort of configuration on that object? 

Thanks.

dsvictor94:
Closes #13 

Limitation: see https://github.com/jlaine/aioice/issues/3
- do not support multiples stun or turn server, the first of each is used
- do not support stuns or turns, it is ignored
- only support udp as transport to turn
- only support "password" credentials to turn

TODO:
- write a exemple
- write tests

### DIFF
--- a/aiortc/__init__.py
+++ b/aiortc/__init__.py
@@ -1,6 +1,6 @@
 from .exceptions import InvalidAccessError, InvalidStateError  # noqa
 from .mediastreams import AudioStreamTrack, VideoStreamTrack  # noqa
-from .rtcconfiguration import RTCConfiguration  # noqa
+from .rtcconfiguration import RTCConfiguration, RTCIceServer  # noqa
 from .rtcdatachannel import RTCDataChannel, RTCDataChannelParameters  # noqa
 from .rtcdtlstransport import (RTCCertificate, RTCDtlsFingerprint,  # noqa
                                RTCDtlsParameters, RTCDtlsTransport)
--- a/aiortc/rtcconfiguration.py
+++ b/aiortc/rtcconfiguration.py
@@ -4,7 +4,8 @@ import attr
 @attr.s
 class RTCConfiguration:
     bundlePolicy = attr.ib(default='max-compat')
-    iceServers = attr.ib(factory=lambda: [RTCIceServer('stun:stun.l.google.com:19302')])
+    iceServers = attr.ib(
+        default=attr.Factory(lambda: [RTCIceServer('stun:stun.l.google.com:19302')]))
 
 
 @attr.s
--- a/aiortc/rtcconfiguration.py
+++ b/aiortc/rtcconfiguration.py
@@ -4,3 +4,21 @@ import attr
 @attr.s
 class RTCConfiguration:
     bundlePolicy = attr.ib(default='max-compat')
+    iceServers = attr.ib(factory=lambda: [RTCIceServer('stun:stun.l.google.com:19302')])
+
+
+@attr.s
+class RTCIceServer:
+    urls = attr.ib()
+    username = attr.ib(default=None)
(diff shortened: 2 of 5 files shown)


dsvictor94:
I do not understand way test fails on travis.
Its appears to be something with attrs factory param, but locally the test pass

jlaine:
This is very cool!

To make the code easier to test, maybe you can have another helper like your parser which checks whether a parsed stun/turn server is supported. This will reduce the number of "continue" statements.

Also: I'd like to do like browsers and have a default stun server (the Google one like currently), so we need to handle these cases differently:

- "the user did not specify the ice servers" -> use the default stun server

- "the user specified an empty list of servers" -> don't use stun / turn

jlaine:
Hi! Regarding your build failures, the "factory" syntactic sugar was only added in a very recent version of the attrs package (18.1.0). So far we have been using these kind of constructs:

```fingerprints = attr.ib(default=attr.Factory(list))```

I'd suggest for now you use the same construct, i.e.:

```iceServers = attr.ib(default=attr.Factory(lambda: [RTCIceServer('stun:stun.l.google.com:19302')]))```

We can change all of these to `factory=` in a different commit, raising the required version of the `attrs` package.

dsvictor94:
I have change the iceServers attr to use default sintaxe, now the tests pass 

About the default behavior I thing the current implementation match your description 
- If the user do not set pass a RTCConfiguration a default one is used instantiated with google stun server
- If the user pass a RTCConfiguration but do not change the iceServers, the google stun is used
- if the user pass explicitly `None` or `[]` no iceServer is used 

About the place where I did the support check, I agree with you, but unfortunately I have no time to do it now

jlaine:
I'm happy with the code as it stands, all we need are the tests to cover the new code and we should be good to merge!

jlaine:
Hm. I merged that a bit fast, there are a couple of bits I'm not sure I understand : 

- where is RTCIceServer.fromdict supposed to be used?
- should we really be silently ignoring unsupported schemes / credential types / multiple servers?

dsvictor94:
I started the implementation because I need integrates with relay services,
normally this this services has temporary credentials and offer many ices servers alternatives.

the way this services provide the iceServers is with REST api that return a already formatted list of RTCIceServer. Help import servers from third parties is the main use of `RTCIceServer.fromdict`. But now looking more carefully to your code maybe a helper function as `candidate_from_sdp` match better your code style

I ignore unsupported servers silently to increase interoperability, but thinking better, raise at unsupported cases appear the right thing to do. But is important to do it when instantiate RTCIceServer to give the user a change to ignore the raise and keep supported servers.

jlaine:
I have broken out the iceServers to "Connection kwargs" into a separate function which is now fully tested.

I also made the following changes:

- `urls` now behaves according to spec, it can be either a string or a list of strings
- the `RTCIceServer.fromdict` helper is gone, you can just use RTCIceServer(**kwargs)
- I have added API documentation, see:

https://aiortc.readthedocs.io/en/latest/api.html#aiortc.RTCIceServer
https://aiortc.readthedocs.io/en/latest/api.html#aiortc.RTCConfiguration

ibc:
> Also: I'd like to do like browsers and have a default stun server (the Google one like currently), so we need to handle these cases differently:

I don't think that's a good idea. Please, do not use "default" ICE servers if none is given.

> * "the user did not specify the ice servers" -> use the default stun server
> * "the user specified an empty list of servers" -> don't use stun / turn

The latter fails. If `iceServers: []` is given, it crashes. See #266

PR #37, opened 2018-07-14
VPN example

alex-eri:


jlaine:
Isn't there some library we can leverage to keep the tun/tap code to a minimum?

One candidate:

https://pypi.org/project/pytap2/

alex-eri:
I did not found ideal "pytap".
Your candidate has

        ret = os.system(ifconfig_cmd)

so i cant run as non root user.

alex-eri:
vpn blocks after 8 megabit/s =)

jlaine:
Unrelated but I'd love your input: I don't like having to copy around the signaling.py in every example, yet it doesn't feel as though it should live in the aiortc library itself. Any suggestions?

My local version of it has additional signaling methods like a Unix socket for local testing

alex-eri:
It posible to send answer before reading offer?

For example.
I want signaling with HTTP POST or PUT with local description as payload returning remote.
Offerer puts desc, server blocks until Answerer puts desc, then server returns answer to offerer and offer to asnswerer.

alex-eri:
Why do you use lists instead of [Queue](https://docs.python.org/3/library/queue.html#queue.Queue) or [asyncio.Queue](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue)?

jlaine:
OK let's stay focused on the VPN example in this ticket, and open other issues for the signaling

PR #47, opened 2018-07-30
[setup] add package aiortc.contrib

mijime:
I tried examples/videostream-cli. but I fail execute.

```
$ python cli.py offer
  File "cli.py", line 10, in <module>
    from aiortc.contrib.media import frame_from_bgr, frame_to_bgr
ModuleNotFoundError: No module named 'aiortc.contrib'
```

Is it necessary to add contrib to package?

```
$ python --version
Python 3.6.6
```

```
$ pip --version
pip 18.0
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setuptools.setup(
         'Programming Language :: Python :: 3.7',
     ],
     cffi_modules=cffi_modules,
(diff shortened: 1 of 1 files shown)


jlaine:
Ouch thanks for catching this!

Issue #14, opened 2018-05-01
h.264 video


petergerten:
great project!

I wonder what would be required to also support H.264?

jlaine:
We would need bindings to an h.264 codec, probably ffmpeg's libavcodec.

This comes in two parts:

- cffi interface to the library's C code (see src/build_vpx.py for example)

- Python code to handle frames and pass them to the codec (see aiortc/codecs/vpx.py for example)

petergerten:
Thanks. I will take a look at vpx.py.
Would it be possible to just use the encoded frames? I intend to re purpose them without decoding.

jlaine:
Grabbing endoded frames sounds like a legitimate usecase buy doesn't sound h.264 specific. Do you have any comparable APIs we can draw inspiration from?

notedit:
May be we can use pyav   https://github.com/mikeboers/PyAV

jlaine:
Initial support for H.264 has landed in `master` thanks to @dsvictor94. We'll be refining it, but I'd already like to point out the great work done by Victor. This is the best way to get your favorite feature into `aiortc` : write the code and the corresponding tests and I'll be happy to review it!

PR #59, opened 2018-08-28
H264 support

dsvictor94:
Add Support to H264

- [x] Depacketizer
- [x] Decoder
- [x] Encoder
- [x] Packetizer
- [ ] Tests
- [ ] Generalize find codec
- [ ] Improve performance


dsvictor94:
closes #14

aiortc/rtcpeerconnection.py
```diff
@@ -37,6 +37,12 @@ def find_common_codecs(local_codecs, remote_codecs):
     for c in remote_codecs:
         for codec in local_codecs:
             if codec.name == c.name and codec.clockRate == c.clockRate:
+                if codec.name == 'H264':
+                    # FIXME: check according to rfc 3184
+                    if c.parameters.get('profile-level-id') != '42001f':
```
jlaine: Instead of hard-coding values, why not compare to codec.parameters?

aiortc/codecs/h264.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __init__(self):
+        self.codec = CodecContext.create('libx264', 'w')
+        self.codec.width = 320
+        self.codec.height = 240
+        self.codec.pix_fmt = 'yuv420p'
+
+        self.raw = open('raw.264', 'wb')
+
+    def encode(self, frame, force_keyframe=False):
```
jlaine: FYI the "force_keyframe" parameter is used in response to PLI (Picture Loss Indication) for VP8. The mechanism may not be the same for H.264, I haven't checked.

aiortc/codecs/h264.py
```diff
@@ -0,0 +1,138 @@
...
+class H264Encoder:
+    timestamp_increment = 3000
+
+    def __init__(self):
+        self.codec = CodecContext.create('libx264', 'w')
+        self.codec.width = 320
+        self.codec.height = 240
+        self.codec.pix_fmt = 'yuv420p'
+
+        self.raw = open('raw.264', 'wb')
```
jlaine: What's this?

jlaine:
This is very cool @dsvictor94 !

Before going any further I strongly suggest you immediately fix the PEP8 errors and start writing the test for the depacketizer.

examples/server/index.html
```diff
@@ -30,15 +30,15 @@
 
 <h2>Options</h2>
 <div class="option">
-    <input id="use-datachannel" checked="checked" type="checkbox"/>
```
jlaine: Revert this, it's unrelated to H.264

examples/server/index.html
```diff
@@ -30,15 +30,15 @@
...
+    <input id="use-datachannel" type="checkbox"/>
     <label for="use-datachannel">Use datachannel</label>
 </div>
 <div class="option">
-    <input id="use-audio" checked="checked" type="checkbox"/>
+    <input id="use-audio" type="checkbox"/>
```
jlaine: Same as above

examples/server/index.html
```diff
@@ -30,15 +30,15 @@
...
     <label for="use-audio">Use audio</label>
 </div>
 <div class="option">
-    <input id="use-video" type="checkbox"/>
+    <input id="use-video" checked="checked" type="checkbox"/>
```
jlaine: This too. Video is purposefully disabled by default to avoid failures on devices with no video.

examples/server/index.html
```diff
@@ -53,6 +53,11 @@ <h2>Options</h2>
         <option value="green">Green frame</option>
         <option value="none">None</option>
     </select>
+    <select id="video-codec">
+        <option value="default">Default</option>
+        <option value="VP8/90000">VP8/90000</option>
+        <option value="H264/90000" selected>H264/90000</option>
```
jlaine: Let "default" be selected by default

examples/server/client.js
```diff
@@ -150,3 +155,56 @@ function stop() {
...
 }
+
+function sdpFilterCodec(codec, realSpd){
+    var allowed = []
+    var codecRegex = new RegExp('a=rtpmap:([0-9]+) '+escapeRegExp(codec))
+    var videoRegex = new RegExp('(m=video .*?)( ([0-9]+))*\\s*$')
+    
+    var lines = realSpd.split('\n');
+
+    var test = false;
```
jlaine: isVideo would probably be a better name.

examples/server/index.html
dsvictor94: Sorry about that. I changed to test faster and I forgot to revert before committing

examples/server/index.html
dsvictor94: Sorry about that. I changed to debug faster and I forgot to revert before committing

aiortc/codecs/h264.py
dsvictor94: Sorry about that too. I was saving encoded packages to file to check video integrity and forgot to cleanup
and of course, the codec parameters must be set according to frames, not hard-coded

aiortc/rtcpeerconnection.py
dsvictor94: I just hard-code to move forward. But I think it must be generalized, because compare parameters compatibility is codec specific, but I not figure out how to do

aiortc/codecs/h264.py
jlaine: Keep the encoded packages, and use them in your unit tests :)

If you need to store a binary file, can you keep the extension as .bin, since I added a rule to .gitattributes to ensure such files are flagged as binary?

aiortc/rtcpeerconnection.py
jlaine: OK that's fine for now. Do make sure you update the `RTCRtpCodecParametersTest` unit test to cover these new code branches though.

examples/server/index.html
jlaine: No need to apologize, I perfectly understand why you had it in your local copy, just pointing out the commit looked like an oversight :)

jlaine:
A suggestion so that we quickly get to some code which can be merged : maybe focus purely on the packetization / depacketization code + full corresponding unit tests (and nothing else). I'd be willing to merge such a PR as a first step.

jlaine:
It looks as though setup.py needs an update to add "av" to it. I'm a bit concerned by the fact there are no wheels for pyav.

jlaine:
As pyav does not provide any wheels (binary distribution), you will need to update the .travis.yml to install the appropriate libav*-dev Debian packages (and the equivalent for homebrew on OS X).

aiortc/codecs/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .g711 import PcmaDecoder, PcmaEncoder, PcmuDecoder, PcmuEncoder
 from .opus import OpusDecoder, OpusEncoder
 from .vpx import VpxDecoder, VpxEncoder
+from .h264 import H264Decoder, H264Encoder
```
jlaine: Please keep the imports sorted alphabetically, you can use `isort` to help you.

aiortc/codecs/h264.py
```diff
@@ -0,0 +1,127 @@
+from itertools import tee
+from struct import unpack_from
+
+from av.packet import Packet
+from av.video.frame import VideoFrame
+from av.codec.context import CodecContext
+
+from ..contrib.media import frame_from_bgr, frame_to_bgr
```
jlaine: The initial intention of contrib.* was that it not be required, as it pulls in opencv. However, I did add opencv to setup.py so I'm not too sure what to do here.

dsvictor94: This is not required, the avlib already return frames in correct format (yuv420p), I just need figure out how extract the bytes direct

dsvictor94:
Do you know how can I get raw RTP (before encrypt) packages to debug? I think I am packetizing correctly the frames but neither chrome neither firefox can decode the frames.

jlaine:
You could hack RTCDtlsTransport._send_rtp for instance.

dsvictor94:
It is the first working version! tested with chrome and firefox
There are a lot more to do (starting with fix travis build and tests), but maybe already safari (not tested yet)

examples/server/server.py
```diff
@@ -23,7 +23,10 @@ def __init__(self, transform):
         self.transform = transform
 
     async def recv(self):
-        frame = await self.received.get()
+        try:
+            frame = await asyncio.wait_for(self.received.get(), 1)
+        except asyncio.TimeoutError:
```
jlaine: Let's not commit these changes, if you want to send bogus video keep the hacks on your laptop :)

jlaine:
I have added the dependency on pyav in `master`, so could you please rebase on top of `master`?

This will give travis a chance to run the test suite. Also, can you please pick up the tests I started writing on the `h264` branch?

https://github.com/jlaine/aiortc/tree/h264

aiortc/rtcrtpsender.py
```diff
@@ -192,7 +193,12 @@ def _send_keyframe(self):
         loop = asyncio.get_event_loop()
 
         encoder = get_encoder(codec)
-        packet = RtpPacket(payload_type=codec.payloadType)
+        # init randomly as rfc3550 section-5.1
```
jlaine: These changes are all legitimate but unrelated to H.264, a separate PR would have been better.

I've integrate the fixes on `master` so you can remove them from this PR.

jlaine:
OK cool CI now passes! We "just" need to get the test coverage back to 100% :) As I said, I provided some tests to exercise the depacketizer in my h264 branch.

One way to exercise the encoder / decoder is to add a modified version of the RTCPeerConnection tests with a video stream, stripping out the vp8 codec to force h264.

jlaine:
By the way for SDP manipulation : instead of using crazy regexes it might be easier to use MediaDescription to parse the SDP, remove what you want then re-serialize it.

jlaine:
Another approach for testing: send VideoFrames into the encoder, and pass to the output to the decoder. Verify we get the expected number of frames with the right size. Testing the actual content of the frames is trickier, though I imagine a mini colour frame probably comes out similar.

aiortc/codecs/h264.py
```diff
@@ -110,18 +115,147 @@ def parse(self, packet):
 
 
 class H264Encoder:
-    timestamp_increment = 3000
+    timestamp_increment = 6000
```
jlaine: This means 15fps, which is unusual, the defaults I have seen in WebRTC are usually 30fps, no?

aiortc/codecs/h264.py
jlaine: Perfect, much better if the library does it for us.

aiortc/codecs/h264.py
```diff
@@ -0,0 +1,261 @@
...
+
+        buffer = io.BytesIO()
+        self.container = av.open(buffer, format='h264', mode='w')
+        self.stream = self.container.add_stream('libx264', rate=25)
+        self.stream.pix_fmt = 'yuv420p'
+        self.stream.codec_context.options = {'profile': 'baseline', 'level': '31'}
+
+    @staticmethod
+    def _packetize_fu_a(data):
+        avaliable_size = PACKET_MAX - FU_A_HEADER_SIZE
```
jlaine: Typo: available not avaliable

jlaine:
I tried sending an AudioFrame through the encoder, but got an empty list as output, any idea why that might be?

aiortc/codecs/h264.py
dsvictor94: Humm.. I choose this number because is the average difference between two consecutive packages Chrome sent to me. Based on RFC, the correct is the frame time, but to do it VideoFrame must have a timestamp field

dsvictor94:
To be resilient to frame corruption I wrapper the encode/decode logic in try-except, this probability make the encoder fail silent on AudioFrame

jlaine:
Sorry there was a typo in my comment, in meant a VideoFrame. I added some print() statements and I'm not hitting the except branch. Could you please add the tests to get coverage back up to 100% so we can merge the PR and iterate from there?

Concerning the frame rate can we put it to 30fps (both timestamp increment + rate which is passed when constricting self.stream) for consistency with vp8?

dsvictor94:
Unfortunately supporting Safari has been deprived and I can no longer use the time at work on it, but I plan to add tests over the weekend.

dsvictor94:
> I tried sending an VideoFrame through the encoder, but got an empty list as output, any idea why that might be?
The encoder was buffering the frames to better compress the final file. I tuned the encoder to zerolatency and now frame by frame is encoded. This solve the absurd latency visualized in server exemple too.

dsvictor94:
I must write test to decoder and common_codec yet but I could not found the equivalent test to VP8 to use as reference

You can point me where this is tested?

jlaine:
Wow, this is amazing progress @dsvictor94 !

You are right there isn't an explicit unit test for VP8 encoding / decoding, but we get coverage because we test an end-to-end RTCPeerConnection <-> RTCPeerConnection using VP8. My suggestion here would be to simply send the output of the encoder through the decoder (and I'll do the same for VP8). It doesn't guarantee our decoder is correct.. but at least that it's compatible with our encoder!

jlaine:
Ah wait, a better approach might be to use the captured H264 packets I provided, as these were generated by a web browser.

jlaine:
@dsvictor94 I have improved the aiortc.VideoFrame <-> av.VideoFrame conversions in a63c02b8ea22e16104060eb7a91b66a3c47c61fd now we no longer need colorspace conversions.

I am not comfortable with the try/except in in _encode_frame, there is no reason for *encoding* to ever blow up. Am I right in assuming this is their because you don't deal with resolution changes (which can and do happen)?

If so, the easiest is probably to do the same as for the VP8 encoder : wait for the first frame before initializing the encoder. If the resolution changes, discard the encoder and re-create a new one.

PR #90, opened 2018-10-30
[examples] Get datachannel-cli working in python3

rlc2:
When running the datachannel-cli demo in python3, it crashes in
json.loads() due to the readline() returning bytes instead of a string.
The parameter to object_from_string() in CopyAndPasteSignaling.receive()
was modified so it looks more like other calls to object_from_string()
in contrib/signaling.py, I.E. added .decode('utf8') to its parameter.

### DIFF
--- a/aiortc/contrib/signaling.py
+++ b/aiortc/contrib/signaling.py
@@ -63,7 +63,7 @@ class CopyAndPasteSignaling:
         print('-- Please enter a message from remote party --')
         descr_str = await self._reader.readline()
         print()
-        return object_from_string(descr_str)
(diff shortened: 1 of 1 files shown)


rlc2:
The travis build failed the test_copy_and_paste (tests.test_contrib_signaling.SignalingTest) unit test with:
AttributeError: 'str' object has no attribute 'decode'

It looks like MockReader.readline() in tests/test_contrib_signaling.py is returning a string instead of bytes.

rlc2:
After looking at the unit test code, it looks like getting it to work in both Python 2.x and Python 3.x is not trivial.  Going to cancel this pull request.

Issue #129, opened 2019-01-22
RTX packet with empty payload causes DTLS to shutdown


zhiweny1122:
Hi jlaine:
I opened a new issue for the dtls close issue in 2 minutes.
I never find this issue on 0.9.13 and it happened on 0.9.18.
I add the debug info in dtlstransport.py in line 480.
The error info is "FAILED unpack requires a buffer of 2 bytes"
Can you give some comments?
I pasted the log here.
"DEBUG:rtp:receiver(video) < RtpPacket(seq=45129, ts=845146785, marker=1, payload=97, 229 bytes)
DEBUG:rtp:sender(video) > RtpPacket(seq=54920, ts=4185232608, marker=0, payload=97, 1300 bytes)
DEBUG:rtp:sender(video) > RtpPacket(seq=54921, ts=4185232608, marker=1, payload=97, 1125 bytes)
DEBUG:rtp:receiver(video) < RtcpSrPacket(ssrc=2859732783, sender_info=RtcpSenderInfo(ntp_timestamp=16136650649041527701, rtp_timestamp=845150655, packet_count=78213, octet_count=86404907), reports=[])
DEBUG:rtp:receiver(video) < RtpPacket(seq=45130, ts=845149665, marker=1, payload=97, 353 bytes)
DEBUG:rtp:receiver(video) > RtcpPsfbPacket(fmt=15, ssrc=3015268611, media_ssrc=0, fci=b'REMB\x02\x00\xaa_\xaat\x0f/\x8b\xdd;i')
DEBUG:rtp:sender(video) > RtpPacket(seq=54922, ts=4185235608, marker=0, payload=97, 1300 bytes)
DEBUG:rtp:sender(video) > RtpPacket(seq=54923, ts=4185235608, marker=1, payload=97, 327 bytes)
DEBUG:rtp:sender(video) > RtpPacket(seq=54924, ts=4185238608, marker=0, payload=97, 1300 bytes)
DEBUG:rtp:sender(video) > RtpPacket(seq=54925, ts=4185238608, marker=1, payload=97, 1006 bytes)
DEBUG:rtp:receiver(video) < RtpPacket(seq=45131, ts=845152905, marker=1, payload=97, 378 bytes)
DEBUG:rtp:receiver(video) > RtcpPsfbPacket(fmt=15, ssrc=3015268611, media_ssrc=0, fci=b'REMB\x02\x00\xa2u\xaat\x0f/\x8b\xdd;i')
[...]

PR #130, opened 2019-01-22
Throw away empty rtp packet

zhiweny1122:
Hi jlaine:
It looks like the issue 129 is related to the rtp empy packet.
fix issue #129


jlaine:
I am closing this as a better fix was committed in 5e6961083e9494cbde5b000a6355f2dab0d080d0

PR #136, opened 2019-01-24
Add options to disable encoding/decoding

quentinwendegass:
Adds a class variable in RTCRtpSender/RTCRtpReceiver to disable encoding/decoding audio frames.

### DIFF
--- a/aiortc/rtcrtpreceiver.py
+++ b/aiortc/rtcrtpreceiver.py
@@ -38,13 +38,17 @@ def decoder_worker(loop, input_q, output_q):
             break
         codec, encoded_frame = task
 
-        if codec.name != codec_name:
-            decoder = get_decoder(codec)
-            codec_name = codec.name
+        if not RTCRtpReceiver.disableDecoding:
+            if codec.name != codec_name:
+                decoder = get_decoder(codec)
+                codec_name = codec.name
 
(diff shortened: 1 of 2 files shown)


koenvo:
I think the `decoder_worker` should not be started at all when decode is disabled.
  ```python
        # if we have a complete encoded frame, handle it
        if encoded_frame is not None:
            encoded_frame.timestamp = self.__timestamp_mapper.map(encoded_frame.timestamp)
            if something.decode_disabled:
              self._track._queue.put(encoded_frame)
            elif self.__decoder_thread:  
              self.__decoder_queue.put((codec, encoded_frame))
```

quentinwendegass:
I took your advice and made some changes. Maybe for now we can just make the options valid if PCMA/PCMU codec is used, so the API keeps being symmetrical? I don't really fully understand the library at this point to come up with a better solution... 

About the global options: In my opinion, this is not a bad approach, but I am happy to take note of a better one. 

About the tests: I'm not really sure how to implement a test for that with your setup. Maybe you can guide me in the right direction here.

263e63a1990904cd8d0bf8fb4cc2903de0091538

jlaine:
Concerning tests: just look in the `tests` directory, it's a collection of vanilla `unittest` tests, very standard.

Issue #150, opened 2019-03-11
Data channel with negotiated = True and custom id


davidskuza:
Hi, 

am I right that we can't use option _negotiated_ and set our custom _id_ (https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel#RTCDataChannelInit_dictionary) on data channel? I see I could pass id to _RTCDataChannel_ instead of calling _createDataChannel_ on _RTCPeerConnection_ but then I'm "not able" to ___createSctpTransport()_. Anyways setting id alone probably doesn't give me much. I need this for already created web "client" which uses this options.

jlaine:
You are right, at the moment negotiated data channels are not supported. I shouldn't be too hard to implement though, and I would certainly review a PR adding this (don't forget to add unit tests).

jlaine:
Closed thanks to David's great work!

PR #151, opened 2019-03-13
Support negotiation out-of-band and custom channel IDs

davidskuza:
Hello,

this PR implements #150. I propose to change initializer of RTCDataChannel, moving argument id to RTCDataChannelParameters as it was not document anyway and it makes more sense to have it there.

This handles two new cases: negotiation out-of-band when we need to provide an id and providing custom id while keeping negotiation in-band.

### DIFF
--- a/aiortc/rtcdatachannel.py
+++ b/aiortc/rtcdatachannel.py
@@ -17,17 +17,28 @@ class RTCDataChannel(EventEmitter):
     :param: parameters: An :class:`RTCDataChannelParameters`.
     """
 
-    def __init__(self, transport, parameters, id=None):
+    def __init__(self, transport, parameters, send_open=True):
         super().__init__()
         self.__bufferedAmount = 0
         self.__bufferedAmountLowThreshold = 0
-        self.__id = id
+        self.__id = parameters.id
         self.__parameters = parameters
         self.__readyState = 'connecting'
         self.__transport = transport
+        self.__send_open = send_open
 
-        if self.__id is None:
-            self.__transport._data_channel_open(self)
+        if self.__parameters.negotiated and (
+            self.__id is None or self.__id < 0 or self.__id > 65534
+        ):
+            raise ValueError('ID must be in range 0-65534 '
+                             'if data channel is negotiated out-of-band')
+
+        if not self.__parameters.negotiated:
+            if self.__send_open:
+                self.__send_open = False
+                self.__transport._data_channel_open(self)
+        else:
+            self.__transport._data_channel_add_negotiated(self)
 
     @property
     def bufferedAmount(self):
@@ -48,6 +59,13 @@ class RTCDataChannel(EventEmitter):
         if value < 0 or value > 4294967295:
             raise ValueError('bufferedAmountLowThreshold must be in range 0 - 4294967295')
         self.__bufferedAmountLowThreshold = value
+
+    @property
+    def negotiated(self):
+        """
+        Whether data channel was negotiated out-of-band.
+        """
+        return self.__parameters.negotiated
 
     @property
     def id(self):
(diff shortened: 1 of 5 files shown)


davidskuza:
I guess you need to look into this travis build configuration. I doubt I should remove those unused import statements...

jlaine:
My guess is that the flake8 version changed since the last build, I'll look into it

aiortc/rtcdatachannel.py
```diff
@@ -17,17 +17,26 @@ class RTCDataChannel(EventEmitter):
...
     """
 
-    def __init__(self, transport, parameters, id=None):
+    def __init__(self, transport, parameters, send_open=True):
         super().__init__()
         self.__bufferedAmount = 0
         self.__bufferedAmountLowThreshold = 0
-        self.__id = id
+        self.__id = parameters.id
+        self.__negotiated = parameters.negotiated
```
jlaine: This isn't useful, as we already store parameters. Let's stay consistent.

aiortc/rtcsctptransport.py
```diff
@@ -1557,7 +1577,8 @@ def _data_channel_open(self, channel):
                 assert stream_id not in self._data_channels
 
                 # one side should be using even IDs, the other odd IDs
-                assert (stream_id % 2) != (self._data_channel_id % 2)
+                # this assert is only valid when not using custom IDs
```
jlaine: Don't keep commented code. If this code is not longer valid, remove it.

aiortc/rtcpeerconnection.py
```diff
@@ -458,7 +458,9 @@ def createDataChannel(self, label, maxPacketLifeTime=None, maxRetransmits=None,
             maxPacketLifeTime=maxPacketLifeTime,
             maxRetransmits=maxRetransmits,
             ordered=ordered,
-            protocol=protocol)
+            protocol=protocol,
+            negotiated=negotiated,
+            id=id)
```
jlaine: Nit: can we keep parameters in alphabetical order please?

aiortc/rtcdatachannel.py
davidskuza: What do you mean? To use parameters directly in code? Should I also remove self.__id?

jlaine: Hm I was tired when I wrote it, keep it as is

jlaine:
I've fixed the isort issues on master, can you rebase your changes so we get a full CI run?

Really looking forward to being able to merge your work! Thanks so much for the new unit tests.

jlaine:
Can you also back out your unrelated change to `aiortc/codecs/__init__.py` ?

davidskuza:
> Can you also back out your unrelated change to `aiortc/codecs/__init__.py` and squash this all down into one commit?

Sorry, missed that one. Should be ok now.

aiortc/rtcdatachannel.py
```diff
@@ -17,17 +17,28 @@ class RTCDataChannel(EventEmitter):
...
         self.__bufferedAmount = 0
         self.__bufferedAmountLowThreshold = 0
-        self.__id = id
+        self.__id = parameters.id
         self.__parameters = parameters
         self.__readyState = 'connecting'
         self.__transport = transport
+        self.__send_open = send_open
+
+        if self.__parameters.negotiated and (
```
jlaine: I hadn't spotted this, but yes, absolutely yes! Thanks for being so thorough.

jlaine:
A big thank you David!

davidskuza:
Well thank you for maintaining such a great project!

I wanted to add .idea directory, created by PyCharm/JetBrains IDEs. Equivalent of .vscode.

aiortc/rtcdatachannel.py
davidskuza: As I am thinking now _self.__send_open_ is useless too. No need to store it.

jlaine:
OK I've update .gitignore in c02c5c13e7106eabd853dbe54cb5712647c418c7

I've also tagged 0.9.20 with your contribution, it will make its way to pypi as soon as CI completes.

PR #155, opened 2019-03-19
Use coroutine instead thread as decoder worker

KarabasYurii:


KarabasYurii:
It doesn't block event loop because frames decoded using ThreadPoolExecutor. Every RTCRtpReceiver create thread for decoder, bigger part of time decoder waiting for input frame. I think it's better to use task in this case. 

If you disagree with my changes, I will close this PR.

jlaine:
I don't necessarily disagree I just fail to see the problem we are solving. Have you run some tests to see how this impacts latency?

One thing definitely looks wrong: asyncio's Queue class is not threads are, so you should not use them for communication between threads.

Also I don't understand why we would use a thread pool to run *long running threads*. A thread pool has a limited size so what happens once you reach the pool limit? Well, new RTP receivers will never decode received media

KarabasYurii:
I think, we have some misunderstanding.

First of all, I don't use asyncio's Queue class to communicate between threads.
`RTCRtpReceiver._handle_rtp_packet` coroutine recv packet and put it into asyncio's Queue, `decoder_worker` coroutine recv this packet through queue and decode it using loop `run_in_executor` method in order to avoid loop blocking.

The main problem I'm trying to solve is to avoid necessary Threads creation, because most of time this threads waiting for `input_q.get()` result. `decoder_worker` is mixin of CPU bound and I/O bound operation, so I am trying to perfome all I/O bound operation with asyncio and CPU bound operation with threading using `ThreadPoolExecutor`.

My implemantion of `decoder_worker` is:
```python
async def decoder_worker(input_q, output_q):
    codec_name = None
    decoder = None

    while True:
        task = await input_q.get()
        if task is None:
            # inform the track that is has ended
            await output_q.put(task)
            break
        codec, encoded_frame = task

        if codec.name != codec_name:
            decoder = get_decoder(codec)
            codec_name = codec.name

        for frame in await asyncio.get_event_loop().run_in_executor(
                None, decoder.decode, encoded_frame
        ):
            # pass the decoded frame to the track
            await output_q.put(frame)

    if decoder is not None:
        del decoder
```


I didn't run specific tests to see how this impacts latency.
I compared **python3.6** builds from **Travis-CI**.

In my case total [test ran](https://travis-ci.com/aiortc/aiortc/jobs/186068318) time is:
```
Ran 394 tests in 162.383s
```
Last `master` [test ran](https://travis-ci.com/aiortc/aiortc/jobs/185474596) took:
```
Ran 394 tests in 163.751s
```

jlaine:
The test suite is not a meaningful test as only few of the tests actually perform any media decoding. Please show me some numbers which prove there is a gain here.

Issue #154, opened 2019-03-19
Add a Websocket-based signaling method for examples


ryogrid:
aiortc datachannel-filexfer example is very very nice example for developers interested in WebRTC and needs reliable p2p direct communication with udp hole punching (data channel is over SCTP and reliable transport) !!!.
But it can not run on Windows platform due to asyncio restriction of pipe.
So I implemented datachannel-filexfer example using websocket signaling server.
(signaling server is copy of code on aioice example )

https://github.com/ryogrid/aiortc_filexfer_sample_for_win32/commit/e69c055abe21c3abf9f5913ab0dca609c9e23a5b

if it has problem for pull request, please comment me :-)
(if this kind post to issues is bad action, sorry...)

ryogrid:
sorry. required for trying is not inough.

[Tested environment]
Python 3.7.1
https://www.python.org/downloads/windows/
Windows x86-64 executable installer

Windows10 Home 64bit version 1803

[aiortc installation]
- for install to windows platform, not normal procedure is required.
- commads below is executed on command prompt

$ git clone https://github.com/ryogrid/aiortc_filexfer_sample_for_win32.git
$ cd aiortc_filexfer_sample_for_win32
$ python setup.py install
$ pip install websockets

ryogrid:
for windows environment installation and running, some modifications are needed...
I forgot these...

https://github.com/ryogrid/aiortc_filexfer_sample_for_win32/compare/e69c05...ee8355

ryogrid:
with mature deliberation, I understand that my modification can not be merge to aiortc.
so I decided create fork of aiortc as library for p2p direct sctp communication.

https://github.com/ryogrid/punch_sctp

sorry for bothering you.

jlaine:
I don't quite understand what happened here, why do you need to fork aiortc at all?

a/ I'm open to providing a websocket-based signaling for the examples

b/ even if I wasn't there is no need to actually fork aiortc to use your own signaling method

ryogrid:
I need functionality of establish p2p ditect data channel only and I want offer it to windows users.
but it cant be realized current aiortc code base (code around media device and codec cant be work on windows platform and making these runnable is not easy. and current pip package could not be installed windows platform due to build error of native bindings...)
so, i decided fork and remove codes not need for data channel and codes not runnable on windows...

i am sorry for code use of aiortc as part of different name library orz

lgrahl:
[...]

PR #169, opened 2019-04-26
add means for data channel use only installation (which can work on windows also)

ryogrid:
pull request responed to #154

This modification enables user to install aiortc (data channel feature only) to windows environment and 
 
[Install procedure]
Because, switch dependent packages from pip.
I add a special version of setup.py (setup_dc.py)
I suppose procedure like below.

```
$ git clone https://github.com/aiortc/aiortc.git --depth 1  
$ cd aiortc  
$ python setup_dc.py install  
````

Due to same package name, data channel only version of aiortc can't coexist with normal version of it.
If user want coexistance, user should use tool like virtualenv.
  
  
[Additional codes to switch data channel feature only mode]
User should insert this codes on top of his or her startup script.

```
import os  
os.environ['AIORTC_SPECIAL_MODE'] = 'DC_ONLY'
```
  
this codes need for avoiding importing of aiortc codes which can't work on data channel feature only installed environment.
  
On code tree, I added check statements using os.environ['AIORTC_SPECIAL_MODE'] on import part.
please check diff.

### DIFF
--- a/aiortc/__init__.py
+++ b/aiortc/__init__.py
@@ -1,7 +1,10 @@
 # flake8: noqa
 
+import os
+
 from .exceptions import InvalidAccessError, InvalidStateError
-from .mediastreams import AudioStreamTrack, MediaStreamTrack, VideoStreamTrack
+if os.getenv('AIORTC_MODE') != 'DC_ONLY':
+	from .mediastreams import AudioStreamTrack, MediaStreamTrack, VideoStreamTrack
 from .rtcconfiguration import RTCConfiguration, RTCIceServer
 from .rtcdatachannel import RTCDataChannel, RTCDataChannelParameters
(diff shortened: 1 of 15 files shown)


jlaine:
Please do not mix unrelated changes withing a single PR or it has no chance of getting merged: one PR for data-channel only, one PR for the new signaling method. The new signaling method does not need a dedicated example, just follow the pattern of the other signaling methods:

- add a class for the new signaling method in aiortc.contrib.signaling
- update `add_signaling_arguments` and `create_signaling`, that way all examples will be able to make use of your new signaling method.

ryogrid:
thanks for comment @jlaine.
i understand. 
i split my modificatin to multi PRs :-)

ryogrid:
because I splited this PR to two PRs respond to a comment fron @jlaine .
#171 
#172

I close this PR.

PR #171, opened 2019-04-26
[retry] add means for data channel use only installation (which can work on windows also)

ryogrid:
this PR is splited modification of https://github.com/aiortc/aiortc/pull/169

---
pull request responed to #154

This modification enables user to install aiortc (data channel feature only) to windows environment and 
 
[Install procedure]
Because, switch dependent packages from pip.
I add a special version of setup.py (setup_dc.py)
I suppose procedure like below.

```
$ git clone https://github.com/aiortc/aiortc.git --depth 1  
$ cd aiortc  
$ python setup_dc.py install  
````

Due to same package name, data channel only version of aiortc can't coexist with normal version of it.
If user want coexistance, user should use tool like virtualenv.
  
  
[Additional codes to switch data channel feature only mode]
User should insert this codes on top of his or her startup script.

```
import os  
os.environ['AIORTC_SPECIAL_MODE'] = 'DC_ONLY'
```
  
this codes need for avoiding importing of aiortc codes which can't work on data channel feature only installed environment.
  
On code tree, I added check statements using os.environ['AIORTC_SPECIAL_MODE'] on import part.
please check diff.

To test 'DC_ONLY' special mode initialization (import codes on code tree),  please use this code.
https://github.com/ryogrid/punch_sctp/blob/02b5f3b3b3f8f60af55c26f78990f5c36f6fe8a9/examples/datachannel-filexfer/import_test.py


jlaine:
I am not comfortable with this `AIORTC_SPECIAL_MODE` environment variable, it feels as though the same could be achieved with just a few `try .. except ImportError`.

ryogrid:
> I am not comfortable with this `AIORTC_SPECIAL_MODE` environment variable, it feels as though the same could be achieved with just a few `try .. except ImportError`.

OK. I understand.

ryogrid:
@jlaine 
>I am not comfortable with this AIORTC_SPECIAL_MODE environment variable, it feels as though the same could be achieved with just a few try .. except ImportError.

I changed implementation according to your comment :-)

jlaine:
There are now binary wheels for Windows of both `aiortc` and `PyAV` in their full form. In light of this I'm not in favour of providing a crippled version of `aiortc`.

PR #172, opened 2019-04-26
add websocket signaling method to contrib.signaling pkg and add websocket version filexfer sample

ryogrid:
this PR is splited modification of https://github.com/aiortc/aiortc/pull/169

---
this PR adds
- websocket signaling method to contrib.signaling
- add websocket version filexfer example code to examples/datachannel-filexfer
- add simple websocket signailing server for above example
- add pip package dependency: websockets(>=7.0) 
  
[Usage]
Please view this.
https://github.com/ryogrid/punch_sctp/blob/5905f37ef59ecfea51ddf8f1558c5ce86247aea7/examples/datachannel-filexfer/README_WS_SIGNALING_VERSION.rst

if user installs aiortc with setup_dc.py (#169 ), below line should be comment-outed.
https://github.com/aiortc/aiortc/compare/master...ryogrid:pr-websocket-version-filexfer?expand=1#diff-9076b3253ed9ef5a2f4eae1ca838df89R12


jlaine:
Why is there a duplicate of the filexfer example? The whole point of having command line arguments to swap the signaling method is to avoid duplication.

You can also remove the text about needing to use websockets on Windows, there was already support for udp + TCP.

We are also going to need unit tests for this new code

examples/datachannel-filexfer/README.rst
```diff
@@ -1,27 +1,37 @@
...
+
 To run the example, you will need instances of the `filexfer` example:
 
 - The first takes on the role of the offerer. It generates an offer which you
   must copy-and-paste to the answerer.
 
 .. code-block:: console
 
-   $ python cli.py send somefile.pdf
+   $ python filexfer.py send somefile.pdf
```
jlaine: This is a good point, I've merged it into master

examples/datachannel-filexfer/README.rst
```diff
@@ -1,27 +1,37 @@
...
 
 - The second takes on the role of the answerer. When given an offer, it will
   generate an answer which you must copy-and-paste to the offerer.
 
 .. code-block:: console
 
-   $ python cli.py receive received.pdf
+   $ python filexfer.py receive received.pdf
```
jlaine: ditto

examples/datachannel-filexfer/filexfer-ws-signaling.py
```diff
@@ -0,0 +1,116 @@
+import argparse
+import asyncio
+import logging
+import time
+import sys
+import os
+
+# from os import path
```
jlaine: No. If you are running aiortc from a git checkout just run from the root of the project:

`pip install -e .`

aiortc/contrib/signaling.py
```diff
@@ -176,15 +177,44 @@ def client_connected(reader, writer):
...
+    async def close(self):
+        if self._websocket is not None and self._websocket.open is True:
+            await self.send(None)
+            await self._websocket.close()
+
+    async def receive(self):
+        try:
+            data = await self._websocket.recv()
+        except asyncio.IncompleteReadError: #TODO: replace to occur from websocket connection
+            print("IncompleteReadError")
```
jlaine: No print statements please

aiortc/contrib/signaling.py
```diff
@@ -176,15 +177,44 @@ def client_connected(reader, writer):
...
         self._writer.write(data + b'\n')
 
+class WebsocketSignaling:
+    def __init__(self, host, port):
+        self._host = host
+        self._port = port
+        self._websocket = None
+
+    async def connect(self):
+        self._websocket = await websockets.connect("ws://" + str(self._host) + ":" + str(self._port))
```
jlaine: This means we don't support wss, I'm starting to wonder whether using urls for the signaling might be a good idea.

ryogrid: this PR supports no encrypted websocket only ok?
or you say that modification which adds support of wss with --signaling-url option?

examples/datachannel-filexfer/README.rst
ryogrid: >You can also remove the text about needing to use websockets on Windows, there was already support for udp + TCP.
there is no description about udp+TCP signaling.
How can i describe about websocket signaling?

like below?
- copy and paste signaling can not work on windows platform
- there is websocket signaling method and it can work at windows platform
- there are other sgnaling methods which can work on windows (but i do not know how to use the mthods to filexfer example...)

examples/datachannel-filexfer/filexfer-ws-signaling.py
ryogrid: ok.

aiortc/contrib/signaling.py
ryogrid: ok.

aiortc/contrib/signaling.py
jlaine: You can keep it like this for now, it was just a comment for future me :)

examples/datachannel-filexfer/README.rst
jlaine: Again, you can control which signaling method is used with command line arguments, so the fact that copy-and-paste doesn't work is irrelevant.

You are right that the additional signaling methods need documentation but here is a start. If sender / receiver are on the same machine all you need is:

**Sender**  (will listen on 127.0.0.1 port 1234 for signaling)

`python filexfer.py -s tcp-socket send somefile.pdf`

**Receiver** (will connect to 127.0.0.1 port 1234 for signaling)

`python filexfer.py -s tcp-socket receive received.pdf`

If the sender / receiver are on different machines, add `--signaling-host <THE IP OF THE SENDER>`.

ryogrid: @jlaine 
I fixed rst files.
but detail of tcp-socket signaling is not described.
I only added the link to above your post.
( I am not good at rst style markup. so I couldn't add above content to README.rst

ryogrid:
I finished modification codes of this PR which respond to review comments from @jlaine.

PR #179, opened 2019-05-31
Add playsinline to video elems for ios compatibility

ortegatron:
This fixes compatibility of html examples with iOS Safari.
Without the playsinline attribute, webkit browsers won't play the video stream, and it will remain black.


jlaine:
Thank you!

PR #185, opened 2019-06-27
v4l2loopback device support for MediaRecorder and janus example

shaunastarabadi:
Added ability for a python janus script to write to a local  v4l2 video device file so that it can be retrieved by other video applications on the system.

### DIFF
--- a/aiortc/contrib/media.py
+++ b/aiortc/contrib/media.py
@@ -333,8 +333,14 @@ class MediaRecorder:
                 codec_name = "aac"
             stream = self.__container.add_stream(codec_name)
         else:
+            print('**MediaRecorder addTrack container.format.name:', self.__container.format.name)
             if self.__container.format.name == "image2":
                 stream = self.__container.add_stream("png", rate=30)
+                stream.pix_fmt = "rgb24"
+            elif self.__container.format.name == "v4l2":
+                # rate=31 workaround for error:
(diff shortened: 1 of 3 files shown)


aiortc/contrib/media.py
```diff
@@ -336,6 +336,11 @@ def addTrack(self, track):
             if self.__container.format.name == "image2":
                 stream = self.__container.add_stream("png", rate=30)
                 stream.pix_fmt = "rgb24"
+            elif self.__container.format.name == "v4l2":
+                # rate=31 workaround for error:
```
jlaine: This framerate looks specific to *your* device?

shaunastarabadi: I suspected the problem is in the ffmpeg library. I couldn't figure out where else to fix it. Any setting of <= 30 fps produces the dts failure, where the video will freeze when it gets to around dts=~1200. By all means, if you can fix it elsewhere, I will be happy to test it in my setup.

The raw format seems to be the only one that works for writing to a device (e.g. /dev/video2 v4l2loopbak), which necessitated supplying a new name to handle it. The else case does not work. Otherwise, the file writes work fine with the else case, as before.
--shaun

jlaine:
Unless you plan to address the issues mentioned above, I'm going to have to close this PR

PR #190, opened 2019-07-10
Move duplicated timestamp logic from AudioStreamTrack and VideoStreamTrack class to separate class

uriyyo:


jlaine:
Thanks for taking the time to put together a PR. I'm kind of torn about what to do with these AudioStreamTrack / VideoStreamTrack classes, as their purpose is not very clear: they are in fact more "DummyAudioStreamTrack" / "DummyVideoStreamTrack", since they are returning dummy frames. A real audio / video stream track could in fact inherit MediaStreamTrack directly, it doesn't necessarily need a next_timestamp method.

jlaine:
On another note : the CI failure seems unrelated to your changes, I'm wondering if homebrew changed the way it package ffmpeg..

uriyyo:
Do I need to rename this classes to "DummyAudioStreamTrack" and "DummyVideoStreamTrack"?

I am currently working on Mac OS and everything work fine, all tests passed successfully.
I was trying to find reason why Mac OS job failed on Travis CI, but didn't find anything. I can try to investigate such behavior on Travis, but I have no experience with it.

aiortc/mediastreams.py
```diff
@@ -7,6 +7,8 @@
 from pyee import EventEmitter
 
 AUDIO_PTIME = 0.020  # 20ms audio packetization
+AUDIO_CLOCK_RATE = 8000
```
jlaine: This kind of bothers me, these are not actually constants, they depend very much on what is negotiated.

Video is different, the clockrate is truly a constant. (VIDEO_PTIME is not..)

uriyyo: This constans used only for AudioStreamTrack, that returns dummy frames. As a suggestion we can rename this const to sothing like `DUMMY_AUDIO_CLOK_RATE`

uriyyo:
Hi @jlaine,

What we will do with a pull request? 
Should I close it?

Thanks

PR #192, opened 2019-07-13
Set correct predicate that skip av module in case when READTHEDOCS set

uriyyo:
Looks like when requirements for "av" module was updated, predicate for install_requires filter wasn't.


jlaine:
Nice catch..

PR #238, opened 2019-12-16
Add type stub file for aioice package

uriyyo:
Add aioice.pyi stub file with type interface for `Candidate` and `Connection` classes from aioice package.

### DIFF
--- a/stubs/aioice.pyi
+++ b/stubs/aioice.pyi
@@ -1,0 +1,65 @@
+from typing import List
+
+
+class Candidate:
+    component: int
+    foundation: str
+    host: str
+    port: int
+    priority: int
+    transport: str
+    related_address: str
+    related_port: int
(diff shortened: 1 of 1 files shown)


jlaine:
Hi and thank you for your PR! As of version 0.6.16, `aioice` should have type hints included in the package. If there are some type hints missing I'd prefer to enhance the type hints directly in `aioice` instead of stubbing them out.

uriyyo:
Great, I will close this pull request.

Also, I am currently trying to add type hints for aiortc modules, but I faced some problems.
Basically I want to add mypy job to Travis CI, but there is a lot of work we must to do.

Should we continue this discussion here or I should create an issue about that?

jlaine:
Let's have a separate issue for adding full typing to aiortc

Issue #234, opened 2019-12-04
Ability to disable pyav


tmatth:
We have maybe a specialized use case for aiortc, in that we use it for stress-testing webrtc servers. Ideally, we'd like to do this without actually encoding/decoding video, and if possible completely bypassing pyav (e.g., just sending dummy packets).

Is this a reasonable goal and would anyone else be interested?

filips123:
WebRTC can also be used for non-video data translfer. For this, requiering `pyav` doesn't make sense, as it is not needed for this.

And as `pyav` requires `FFmpeg` and other system codecs, installation could be difficult, even if you won't actually use them. Because of that, `aiortc` should not require/force packages that are needed "just" for video transfer.

jlaine:
This issue regularly crops up, but I just don't have time to maintain multiple favours of aiortc.

Issue #118, opened 2019-01-11
Allow aiortc to be used as a pure Python module


sametmax:
Currently, installing aiortc fail if you don't have the libavdevice-dev libavfilter-dev libopus-dev libvpx-dev pkg-config modules installed on linux.

However, correct me if I'm wrong, but they are only useful if you wish to use sound or or video. For pure data, they are not required, are they ?

If that's the case, I would suggest to allow the package to be able to be installed in 3 modes:

```bash
pip install aiortc # pure Python, for data only
pip install aiortc[audio] # with audio support
pip install aiortc[video] # with audio and video support
```

This can be done with the "extras" section in setup.py.

Indeed, webrtc is useful already without audio or video, it solves a complex problems with TURN and STUN servers, encryption and a way for 2 peers to connect to each others. And it does not in a standard way, that works out of the box in the web browser.

Hence, any program wanting to exchange data peer 2 peer may want to implement just the data part of webrtc, which can be done without complex media dependencies and be deployed in a bigger set of plateforms (anywhere where the dependency wheels exist).

jlaine:
This is a duplicate of #78

Issue #78, opened 2018-10-08
DataChannel-only 'feature' without codec dependencies?


warner:
Great stuff! I'd love to use this in [magic-wormhole](https://github.com/warner/magic-wormhole).. I only need the DataChannel, though. Any idea how hard it'd be to define a 'feature' for the codecs, so `pip install aiortc[video]` gets you the dependencies on the codecs, but `pip install aiortc[datachannel]` does not?

jlaine:
I'd like the default install to keep pulling in the necessary bits for all of aiortc to work (I don't want to get swamped with "why doesn't media work" tickets), but if you want to put together a PR to have a "no media" variant I'll definitely look at it.

warner:
Cool, thanks, I'll see what I can do. One solution might be to split off a separate `aiortc-datachannel` distribution, then make `aiortc` depend upon it. Might be easier than using features, and would retain the "depend on just `aiortc` to get everything" property.

jlaine:
I'm closing this: we have binary wheels for all of our dependencies on a wide array of platforms.

PR #243, opened 2019-12-27
Add datachannel-only version of `aiortc`

filips123:
This PR adds datachannel-only version of `aiortc` which can work without `cffi`, `av` and FFmpeg. Is is quick fix for #78, #118 and #234, but I wouldn't close this issues yet as this should be handled in some better way (see bellow). This also replaces #171.

I added `datachannel_only` variable to `setup.py` which is by default `False`. However, when publishing with Travis CI, the scripts overwrites it with `sed` to `True`. And when it is `True`, it will remove all `cffi` and `av` dependencies, making it pure-Python module. It will also change package name and description to include `-datachannel`.

After publishing, this will create additional `aiortc-datachannel` package on PyPI, which will be the same as normal `aiortc` package, but without codec dependencies. Users who will want to use `aiortc` just for datachannel will then have to isntall that package instead of default one. The reason for separate package is what @jlaine said in https://github.com/aiortc/aiortc/issues/78#issuecomment-427922168 and I agree with that.

In the source code, there are now a bunch of `try...except` statements around modules that rely on codecs. This is not really nice, but it should work for now. The proper solution would be what @warner said in https://github.com/aiortc/aiortc/issues/78#issuecomment-429406897, but this would require more work and can/should be done in the future.

I partly tested this on Debian Buster and Windows 10 with `datachannel-filexfer`. On Debian, I can run program and it correctly starts and prints message that I should send to remote party. However, I didn't have time to test if actually everything works. But on Windows, this happens:

```
-- Please send this message to the remote party --
{"type": "bye"}

  File "filexfer.py", line 115, in <module>
    loop.run_until_complete(coro)
  File "C:\Program Files\Python37\lib\asyncio\base_events.py", line 568, in run_until_complete
    return future.result()
  File "filexfer.py", line 63, in run_offer
    await signaling.connect()
  File "d:\users\filips\downloads\aiortc\src\aiortc\contrib\signaling.py", line 126, in connect
    lambda: asyncio.StreamReaderProtocol(self._reader), self._read_pipe
  File "C:\Program Files\Python37\lib\asyncio\base_events.py", line 1421, in connect_read_pipe
    transport = self._make_read_pipe_transport(pipe, protocol, waiter)
  File "C:\Program Files\Python37\lib\asyncio\base_events.py", line 433, in _make_read_pipe_transport
    raise NotImplementedError
NotImplementedError

During handling of the above exception, another exception occurred:

  File "filexfer.py", line 121, in <module>
    loop.run_until_complete(signaling.close())
  File "C:\Program Files\Python37\lib\asyncio\base_events.py", line 568, in run_until_complete
    return future.result()
  File "d:\users\filips\downloads\aiortc\src\aiortc\contrib\signaling.py", line 132, in close
    self._read_transport.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

I don't know if this is related to this PR or `aiortc` generally doesn't work on Windows. In any case, it would be good to fix it.


jlaine:
Could you rebase your change on top of the latest master, to see where we stand?

DurandA:
Shouldn't `av` and `cffi` be part of [extras_require](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)? How can we install the module without codecs dependencies without patching `datachannel_only = True` in _setup.py_?

Thank you very much for the PR, I hope this gets merged soon.

filips123:
> Shouldn't av and cffi be part of extras_require?

I think that @jlaine once said (in some other issue or PR) that he would want `aiort` package to be fully-function package with support for everything, but with `extras_require`, people would also have to install those additional dependencies.

I would more like that option, also with better handling than just a bunch of tries around the imports, but it may require more work.

> How can we install the module without codecs dependencies without patching datachannel_only = True in setup.py?

Until new package is released, you would have to manually patch `setup.py`. Once it is released, separate `aiortc-datachannel` package would be published to PyPI which would include patched `setup.py` for just datachannel-only version.

DurandA:
> I think that @jlaine once said (in some other issue or PR) that he would want `aiort` package to be fully-function package with support for everything, but with `extras_require`, people would also have to install those additional dependencies.

With e.g. ` extras_require` specified as below, installing _aiortc_ using `pip install aiortc[media]` will also install `av` and `cffi`.
```
extras_require={
    "media": ["av>=7.0.0,<8.0.0", "cffi>=1.0.0"],
}
```

I just wanted to make sure that you are aware of this option. :-)

jlaine:
> With e.g. ` extras_require` specified as below, installing _aiortc_ using `pip install aiortc[media]` will also install `av` and `cffi` (and `aiortc[media]` can be used as dependency by other projects).
> 
> ```
> extras_require={
>     "media": ["av>=7.0.0,<8.0.0", "cffi>=1.0.0"],
> }
> ```
> 
> I just wanted to make sure that you are aware of this option. :-)

Right, this means that the default install will give you an incomplete WebRTC implementation, so this is what I'd like to avoid.

DurandA:
> Right, this means that the default install will give you an incomplete WebRTC implementation, so this is what I'd like to avoid.

Are there any particular reason or unwanted side-effects for this? I really don't meant to be annoying with this question. :-)

I think other options include [namespace package](https://packaging.python.org/guides/packaging-namespace-packages/) (e.g. splitting everything into `core` and `media` distributions packages) or [creating plugin](https://packaging.python.org/guides/creating-and-discovering-plugins/) `aiortc_media` plugin. With either of this solution, it should be possible to `pip install aiortc` by naming the _media_ package to `aiortc` and the _core_ package to `aiortc-core` or similar. The downside is that it requires major refactoring of `aiortc`.

@jlaine What is your opinion about this? Would you consider a PR with a complete packaging overhaul?

jlaine:
> > Right, this means that the default install will give you an incomplete WebRTC implementation, so this is what I'd like to avoid.
> 
> Are there any particular reason or unwanted side-effects for this? I really don't meant to be annoying with this question. :-)

I'm not annoyed by the question :) My point is that `aiortc` is a WebRTC implementation, it's not mean to be a "non-standard conforming implementation which just implements datachannels". Changing this now means we'll break any project that uses `aiortc`, and will lead to far more support requests than I can possibly handle. I'm not opposed to having a "cut-down" version of aiortc which disables all audio/video support, but this needs to be opt-in and needs to be tested by CI.

Taking a different tack on this topic: what's the actual problem we are trying to solve? How does having support for audio/video harm you? Most likely the answer is something along the lines of:

- "installing PyAV is tricky because it requires FFmpeg"
- "installing aiortc is tricky because it requires opus + libvpx".

The second point (opus / vpx) is almost solved, I'm planning to ship binary wheels for both Linux and OS X which ship with all required libraries. These wheels are already being built by CI at the moment. Ideally I'd also ship Windows wheels but I haven't worked out all the kinks of compiling code on Windows.

The first point (PyAV itself) needs to be solved upstream, but would take the same approach: provide binary wheels for PyAV, containing the FFmpeg libraries.

jlaine:
PyAV 8.0.0 now ships with binary wheels, meaning you won't need the FFmpeg libraries to install aiortc. I'll try and finish my work on binary wheels for aiortc itself, and then I'm not convinced there is still a reason to continue with this PR.

DurandA:
> PyAV 8.0.0 now ships with binary wheels

Do you know if this includes ARM platforms?

jlaine:
> > PyAV 8.0.0 now ships with binary wheels
> Do you know if this includes ARM platforms?

It does not for the moment. Feel free to go submit a PR against https://github.com/jlaine/pyav-ffmpeg/ which is where I produce the FFmpeg binary builds which are then used for PyAV wheels.

jlaine:
OK we now have binary wheels for `aiortc` and all it's dependencies. I'm closing this PR, but feel free to continue the discussion if you have convincing arguments about why we need a datachannel-only version of `aiortc`.

PR #260, opened 2020-02-08
Add libsrtp2-dev to requirements for Debian/Ubuntu

daanzu:
Needed to compile `pylibsrtp`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,7 +91,7 @@ On Debian/Ubuntu run:
 
 .. code:: bash
 
-    apt install libavdevice-dev libavfilter-dev libopus-dev libvpx-dev libsrtp2-dev pkg-config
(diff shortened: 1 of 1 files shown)


jlaine:
This is not actually a hard requirement: I would expect `pylibsrtp` to get installed as a binary wheel, in which case it's already statically linked against libsrtp2. In this case you really do not need libsrtp2-dev.

Have you tried updating pip (`pip install -U pip`) before installing aiortc?

Nevertheless I'm open to merging this change if you could just tweak it slightly to keep the dependencies ordered alphabetically?

daanzu:
Ah, I was installing on a `armv7l` platform, where apparently no binary wheel `pylibsrtp` is available.

I alphabetized.

PR #261, opened 2020-02-09
Support alsa hardware output in MediaRecorder

seler:
This change allows for MediaRecorder to stream tracks to alsa hardware output.

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -325,7 +325,7 @@ class MediaRecorder:
         :param track: A :class:`aiortc.MediaStreamTrack`.
         """
         if track.kind == "audio":
-            if self.__container.format.name == "wav":
+            if self.__container.format.name in ("wav", "alsa"):
                 codec_name = "pcm_s16le"
             elif self.__container.format.name == "mp3":
                 codec_name = "mp3"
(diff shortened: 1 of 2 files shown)


jlaine:
This is an interesting change, do you have a link to a document which shows that this is always the right sample format for alsa?

seler:
Alsa is pcm, same as wav.

seler:
> This is an interesting change, do you have a link to a document which shows that this is always the right sample format for alsa?

Alsa is pcm, same as wav.

seler:
Is it even possible to write a test for alsa without the hardware? Any suggestions?

jlaine:
Does ALSA have a concept of a "null device" we can maybe leverage?

Otherwise I shouldn't sweat it, if it works for you and doesn't trigger a regression, I'll for once have to live without a test :)

seler:
> Does ALSA have a concept of a "null device" we can maybe leverage? In any case I expect the test would need to be disable on any platform that isn't Linux.

```
$ arecord -L
null
    Discard all samples (playback) or generate zero samples (capture)
```
It does. Except you still need alsa set up on server. More info on [alsa wiki](https://www.alsa-project.org/main/index.php/Matrix:Module-dummy)

PR #262, opened 2020-02-10
Replace old-style string formatting with f-strings

uriyyo:
The minimal python version for `aiortc` is 3.6 so we can replace all old-style formatting `%` by using f-strings which available from python 3.6.


jlaine:
@uriyyo This commits seems to have an unwanted side-effect: the logging calls in the SCTP stack result in serializing various objects even when logging_debug is a no-op lambda. This is not great for performance.

uriyyo:
@jlaine looks like I replaced log message formating with f strings, so they immediately convert all arguments. I will fix this bug as soon as possible.

PR #268, opened 2020-02-18
Add mypy job to .travis.yml

uriyyo:
What was done in the scope of this pull request:
1. Rewrite all `type comments` into `type annotations`.
2. Edit code due to mypy errors.

There is a problem with `aiortc.rtcsctptransport:Chunk` class. In this class `body` attribute defined as an instance attribute but in delivered classes it defined using `property` due to it mypy throws an error. In order to suppress errors `# type: ignore` was used on the `body` property declaration.

Also, there is problem with `aiorc.rtcsctptransport:RTCSctpTransport`. `_send` method dynamically add attributes to `DataChunk` instance, due to it mypy throws an error, suppress them using `no_type_check` decorator.

I think we should resolve this problem in the future.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+exclude_lines =
+    pragma: no cover
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ language: python
 matrix:
   include:
   - env: BUILD=lint
-    install: pip install black flake8 isort
+    install: pip install black flake8 isort mypy
     python: "3.6"
   - env: BUILD=test
     language: generic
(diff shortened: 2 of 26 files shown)


jlaine:
This is awesome! I need to check why the test coverage has changed though, this is not expected.

uriyyo:
I found a problem. I exclude body of abstract methods from coverage report, it will increase coverage percent.

jlaine:
NOTE: in the future please keep any settings in `setup.cfg`

jlaine:
I found another reason why the coverage report had changed: the `tests` directory and `setup.py` were being included! The root cause: you added a .coveragerc, which overrode the settings I had put in setup.cfg

uriyyo:
Sorry, my bad. I didn't notice that there is` setup.cfg` with `mypy` configuration.

Issue #270, opened 2020-02-18
Transport closes in setRemoteDescription (renegotiation) if media and datachannel are bundled


jmillan:
Simply that.

Create a bundled session with media and datachannel, renegotiate and transport is closed and never reconnected.

jlaine:
Can we please have an actual unit test for renegotiation?

PR #271, opened 2020-02-18
fix: Do not close transport on 'setRemoteDescription' if media and da…

jmillan:
…ta are bundled (#270)

### DIFF
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -847,7 +847,11 @@ class RTCPeerConnection(AsyncIOEventEmitter):
                     transceiver.sender.setTransport(masterTransport)
                     transceiver._bundled = True
                     transceiver._transport = masterTransport
-            if self.__sctp and self.__sctp.mid in slaveMids:
+            if (
+                self.__sctp
+                and self.__sctp.mid in slaveMids
+                and not self.__sctp._bundled
+            ):
                 oldTransports.add(self.__sctp.transport)
                 self.__sctp.setTransport(masterTransport)
                 self.__sctp._bundled = True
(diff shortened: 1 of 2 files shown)


jlaine:
I'm not sure I understand the issue, could you please provide a test case so we can lock down the behaviour?

jmillan:
Yes,

While in a session with a bundled media track and a datachannel try to, for example, stop sending the media track by doing the corresponding SDP offer answer.

The only transport (because everything is being bundled) will be closed.

jlaine:
OK I'll try to write a unit test for this and run it by you

jmillan:
I was just checking the tests right now. I could write it if you prefer.

jlaine:
Even better!

jmillan:
There you are

jlaine:
Thank you very much for the test, there is just a "print" statement which needs to go :)

Unless I'm mistaken the problem stems from here:

https://github.com/aiortc/aiortc/blob/f7b78b82a4842965c677cccc4f807c84f8e37dd8/src/aiortc/rtcpeerconnection.py#L850

It looks as though we are missing the `and not self.__sctp._bundled` condition to avoid running this code twice, or am I mistaken? If this is correct I'd rather have the following change:

```
diff --git a/src/aiortc/rtcpeerconnection.py b/src/aiortc/rtcpeerconnection.py
index 5d4474a..3a85ab9 100644
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -847,7 +847,11 @@ class RTCPeerConnection(AsyncIOEventEmitter):
                     transceiver.sender.setTransport(masterTransport)
                     transceiver._bundled = True
                     transceiver._transport = masterTransport
-            if self.__sctp and self.__sctp.mid in slaveMids:
+            if (
+                self.__sctp
+                and self.__sctp.mid in slaveMids
+                and not self.__sctp._bundled
+            ):
                 oldTransports.add(self.__sctp.transport)
                 self.__sctp.setTransport(masterTransport)
                 self.__sctp._bundled = True
```

jmillan:
LGTM @jlaine

jlaine:
OK I've updated the PR let's see if CI passes

Issue #272, opened 2020-02-18
Use dataclasses instead of attrs


uriyyo:
Feature request.

Since python 3.7 version `dataclasses` is a part of python standard library. At python 3.6 we can use `dataclasses` as a third-party package.

We can move the existing code base from `attrs` to `dataclasses`.

I can do it by myself.

jlaine:
Yeah, this is one of the benefits of dropping support for Python 3.5 :)

On `aioquic` which has been Python 3.6+ from the start this is what I do:

https://github.com/aiortc/aioquic/blob/ee7fbbe9a4ebf93fb1f009dd59d4a89a5e704639/setup.py#L62

jlaine:
A PR is most welcome. Like your typing changes, this is something I had on my radar but never found time for.

The one area I'd like a micro benchmark is for the SCTP chunk classes, especially DATA and SACK as they are on the hot path.

PR #273, opened 2020-02-19
Use dataclasses instead of attrs

uriyyo:
Close: #272

### DIFF
--- a/src/aiortc/rtcrtpparameters.py
+++ b/src/aiortc/rtcrtpparameters.py
@@ -1,23 +1,22 @@
 from collections import OrderedDict
-from typing import List  # noqa
+from dataclasses import dataclass, field
+from typing import List, Optional  # noqa
 
-import attr
 
-
-@attr.s
+@dataclass
 class RTCRtpCodecCapability:
     """
     The :class:`RTCRtpCodecCapability` dictionary provides information on
     codec capabilities.
     """
 
-    mimeType: str = attr.ib()
+    mimeType: str
     "The codec MIME media type/subtype, for instance `'audio/PCMU'`."
-    clockRate: int = attr.ib()
+    clockRate: int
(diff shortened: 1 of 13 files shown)


src/aiortc/rtcrtpparameters.py
```diff
@@ -1,48 +1,47 @@
 from collections import OrderedDict
-from typing import List  # noqa
+from dataclasses import dataclass, field
+from typing import List, Optional  # noqa
```
jlaine: The "noqa" and any similar ones can go. The reason it was there was to silence flake8 since the import wasn't actually referenced in code.

src/aiortc/rtcrtpparameters.py
```diff
@@ -56,116 +55,114 @@ def __str__(self):
...
 
-@attr.s
+@dataclass
 class RTCRtpCodingParameters:
-    ssrc: int = attr.ib()
-    payloadType: int = attr.ib()
-    rtx: RTCRtpRtxParameters = attr.ib(default=None)
+    ssrc: int
+    payloadType: int
+    rtx: RTCRtpRtxParameters = None
```
jlaine: Optional[RTCRtpRtxParameters] ?

src/aiortc/rtcrtpparameters.py
uriyyo: Remove "noqa" comment, because of `List` and `Optional` really used in this module.

src/aiortc/rtcrtpparameters.py
uriyyo: Set type to `Optional[RTCRtpRtxParameters]`.
We are using `strict_optional = False` configuration in `setup.cfg` so mypy doesn't check for None type cases.

PR #286, opened 2020-02-22
Fixes #283

### LINKED ISSUE #283: await pc.close() does not complete if its sending tracks were stopped() before
Our Python app (which uses `aiortc`) uses the default asyncio `loop`:

```py
loop = asyncio.get_event_loop()
```

I expect this to be the same `loop` used by `aiortc` code. When we wish to finish our Python program we close all our asyncio tasks and so on, and also close all `aiortc` objects:

```py
    async def shutdown() -> None:
        # close our channel
        await channel.close()

        # close all aiortc MediaPlayers
        for player in players.values():
            if player.audio:
                player.audio.stop()
            if player.video:
                player.video.stop()

        # close all aiortc PeerConnections
        for pc in pcs.values():
            await pc.close()

        # stop the loop (just in case)
        loop.stop()
```

```py
    try:
        loop.run_until_complete(
            # our stuff here
        )
    # reached after calling channel closure
    except RuntimeError:
        pass
    finally:
        loop.run_until_complete(
            shutdown()
        )
```

The thing is, when calling `shutdown()` function, sometimes the python process does **NOT** terminate. By adding some logs we **clearly** see that, eventually, a call to `await pc.close()` **never** completes, and that's the problem.

This usually happens when the webcam is active.

**However**, if we close all the pcs **before** we close the `players`, then everything works fine. So IMHO there is a bug when closing a `MediaPlayer` whose tracks are being used by a `PeerConnection`.

This is easily reproducible. We can also just close a specific `MediaPlayer` and, if its tracks were being sent by a `PeerConnection`, then `await pc.close()` does **NOT** complete.

jmillan:
- MediaPlayer: close av container when there are no remaining tracks
 - RTCRtpSender: don't close MediaStreamTrack on close()
 - RTCRtpSender: close() stops sending RTP and RTCP

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -280,10 +280,9 @@ class MediaPlayer:
             self.__thread_quit.set()
             self.__thread.join()
             self.__thread = None
-
-        if self.__container:
-            self.__container.close()
-            self.__container = None
+            if self.__container:
+                self.__container.close()
+                self.__container = None
 
     def __log_debug(self, msg, *args):
         logger.debug(f"player(%s) {msg}", self.__container.name, *args)
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -193,7 +193,8 @@ class RTCRtpSender:
             self.__transport._unregister_rtp_sender(self)
             self.__rtp_task.cancel()
             self.__rtcp_task.cancel()
-            await asyncio.gather(self.__rtp_exited.wait(), self.__rtcp_exited.wait())
+            self.__rtp_exited.set()
+            self.__rtcp_exited.set()
 
     async def _handle_rtcp_packet(self, packet):
         if isinstance(packet, (RtcpRrPacket, RtcpSrPacket)):
@@ -286,7 +287,7 @@ class RTCRtpSender:
         sequence_number = random16()
         timestamp_origin = random32()
         try:
-            while True:
+            while self.__rtp_exited.is_set is False:
                 if not self.__track:
                     await asyncio.sleep(0.02)
                     continue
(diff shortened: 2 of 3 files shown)


jmillan:
@jlaine,

This may not be perfect as I don't have enough internal knowledge to say this won't break anything.

Of course the fact that `RTCRtpSender.stop()` does not close the `MediaStreamTrack` changes the API.

jmillan:
mmm, this has some drawbacks..

````
FAIL: test_connect_video_bidirectional (tests.test_rtcpeerconnection.RTCPeerConnectionTest)
----------------------------------------------------------------------
  File "/Users/jmillan/src/aiortc/tests/test_rtcpeerconnection.py", line 2275, in test_connect_video_bidirectional
    "transport",
AssertionError: Lists differ: ['outbound-rtp', 'transport'] != ['inbound-rtp', 'outbound-rtp', 'remote-inb[41 chars]ort']

First differing element 0:
'outbound-rtp'
'inbound-rtp'

Second list contains 3 additional elements.
First extra element 2:
'remote-inbound-rtp'

- ['outbound-rtp', 'transport']
+ ['inbound-rtp',
+  'outbound-rtp',
+  'remote-inbound-rtp',
+  'remote-outbound-rtp',
+  'transport']

----------------------------------------------------------------------
Ran 1 test in 5.739s

FAILED (failures=1)
````

jmillan:
fixed...

jlaine:
I think we are going to need a documentation section explaining the lifetime of tracks, because I honestly don't understand it anymore.

I understand the current code was not aligned with specs but I had out in some effort to avoid memory leaks, this is something that's going to need revisiting.

In any case many thanks for the PR it helps me understand what we are aiming for.

jmillan:
The concept is very simple. A local MediaStreamTrack is independent from RTCPeerConnection. It is sourced by a media device, file or URL. It can be attached to a RTCPeerConnection or it may not.

When a local MediaStreamTrack is added to a RTCPeerConnection it simply creates a RTCTransceiver with the corresponding RTCSender which will be responsible of sending the media generated by the track through the transport.

When a local MediaStreamTrack is removed from a RTCPeerConnection, the corresponding RTCSender will stop sending any media generated by the track.

So the life time of a local MediaStreamTrack is independent to the RTCPeerConnection. The later will just send its media or stop sending when indicated. It will not create, nor stop it. Creating and stopping local MediaStreamTrack is user responsibility.

Issue #237, opened 2019-12-12
Segmentation fault since MediaPlayer cleanup change


tmatth:
We use aiortc for testing Janus.

Since this commit:
https://github.com/aiortc/aiortc/commit/0310cd8486b38d8d8f295650253224247c43ed42

We are seeing the following segmentation fault (the backtrace might not be very useful):
```
Fatal Python error: Segmentation fault

Thread 0x00007f14697fa700 (most recent call first):
  File "/usr/lib/python3.6/threading.py", line 295 in wait
  File "/usr/lib/python3.6/queue.py", line 164 in get
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 67 in _worker
  File "/usr/lib/python3.6/threading.py", line 864 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/lib/python3.6/threading.py", line 884 in _bootstrap

Thread 0x00007f1469ffb700 (most recent call first):
  File "/usr/lib/python3.6/threading.py", line 295 in wait
  File "/usr/lib/python3.6/queue.py", line 164 in get
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 67 in _worker
  File "/usr/lib/python3.6/threading.py", line 864 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/lib/python3.6/threading.py", line 884 in _bootstrap

Thread 0x00007f146a7fc700 (most recent call first):
  File "/usr/lib/python3.6/threading.py", line 295 in wait
  File "/usr/lib/python3.6/queue.py", line 164 in get
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 67 in _worker
  File "/usr/lib/python3.6/threading.py", line 864 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
  File "/usr/lib/python3.6/threading.py", line 884 in _bootstrap

Thread 0x00007f146affd700 (most recent call first):
  File "/usr/lib/python3.6/threading.py", line 295 in wait
  File "/usr/lib/python3.6/queue.py", line 164 in get
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 67 in _worker
  File "/usr/lib/python3.6/threading.py", line 864 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
[...]

PR #287, opened 2020-02-22
[media player] only close container once all tracks are stopped

jlaine:
When reading a file with multiple tracks, we must take care to only
close the container once the player thread has shut down. This thread is
only shut down once all tracks are stopped.

Fixes: #237, #274


jmillan:
:+1: Yeah

PR #292, opened 2020-02-24
Mediaplayer deadlock fix

jmillan:
* MediaPlayer deadlock fix when MediaStreamTrack is stopped before stopping the `RTCRtpSender` to which the track is attached
* Don't stop track on `RTCRtpSender.stop()`

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -156,6 +156,7 @@ class PlayerStreamTrack(MediaStreamTrack):
         super().__init__()
         self.kind = kind
         self._player = player
+        self._throttle_playback = player._throttle_playback
         self._queue = asyncio.Queue()
         self._start = None
 
@@ -171,7 +172,7 @@ class PlayerStreamTrack(MediaStreamTrack):
         frame_time = frame.time
 
         # control playback rate
-        if self._player._throttle_playback and frame_time is not None:
+        if self._player is not None and self._throttle_playback and frame_time is not None:
             if self._start is None:
                 self._start = time.time() - frame_time
             else:
@@ -223,6 +224,10 @@ class MediaPlayer:
         self.__thread: Optional[threading.Thread] = None
         self.__thread_quit: Optional[threading.Event] = None
 
+        # check whether we need to throttle playback
+        container_format = set(self.__container.format.name.split(","))
+        self._throttle_playback = not container_format.intersection(REAL_TIME_FORMATS)
+
         # examine streams
         self.__started: Set[PlayerStreamTrack] = set()
         self.__streams = []
@@ -235,10 +240,6 @@ class MediaPlayer:
             elif stream.type == "video" and not self.__video:
                 self.__video = PlayerStreamTrack(self, kind="video")
                 self.__streams.append(stream)
-
-        # check whether we need to throttle playback
-        container_format = set(self.__container.format.name.split(","))
-        self._throttle_playback = not container_format.intersection(REAL_TIME_FORMATS)
 
     @property
     def audio(self) -> MediaStreamTrack:
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -326,9 +326,8 @@ class RTCRtpSender:
         except (asyncio.CancelledError, ConnectionError, MediaStreamError):
             pass
 
-        # stop track
+        # unreference the track
         if self.__track:
-            self.__track.stop()
             self.__track = None
 
         self.__log_debug("- RTP finished")
--- a/tests/test_rtcpeerconnection.py
+++ b/tests/test_rtcpeerconnection.py
@@ -1,5 +1,6 @@
 import asyncio
 import re
+import platform
 from unittest import TestCase
 
 import aioice.stun
@@ -1790,17 +1791,12 @@ a=rtpmap:0 PCMU/8000
             ["stable", "have-remote-offer", "stable", "closed"],
         )
 
-    def _test_connect_audio_and_video_mediaplayer(self, stop_tracks: bool):
+    def _test_connect_audio_and_video_mediaplayer(self, track_factory, stop_tracks: bool):
         """
         Negotiate bidirectional audio + video, with one party reading media from a file.
 
         We can optionally stop the media tracks before closing the peer connections.
         """
-        media_test = MediaTestCase()
-        media_test.setUp()
-        media_path = media_test.create_audio_and_video_file(name="test.mp4", duration=5)
-        player = MediaPlayer(media_path)
-
         pc1 = RTCPeerConnection()
         pc1_states = track_states(pc1)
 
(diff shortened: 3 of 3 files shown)


src/aiortc/contrib/media.py
```diff
@@ -171,7 +172,7 @@ def __init__(self, player, kind):
         frame_time = frame.time
 
         # control playback rate
-        if self._player._throttle_playback and frame_time is not None:
```
jmillan: Deadlock was happening here. At the time the first expression in the condition is being checked by a thread, another thread is closing the player, due to its tracks closure and it would become `None`.

src/aiortc/rtcrtpsender.py
```diff
@@ -326,9 +326,8 @@ def _send_keyframe(self) -> None:
         except (asyncio.CancelledError, ConnectionError, MediaStreamError):
             pass
 
-        # stop track
+        # unreference the track
         if self.__track:
-            self.__track.stop()
             self.__track = None
```
ibc: Actually no need to check it. Just do `self.__track = None` without any previous check.

ibc:
Amazing.

src/aiortc/contrib/media.py
jlaine: I'm not sure I understand this, there should be no threading involved here: everything happens on the main thread, except for `player_worker` which does not touch the `Player` or `PlayerStreamTrack` objects.

jmillan: @jlaine,

I'm not sure  I understand it neither. Flow is blocked in such line, just put a log before and after it. 

git checkout 2b4fce4b58aa3c56a05a112ab4d0634e0c45a3a6 in the PR and run the new tests added in such commit (those retrieving the system webcam/mic)

Hope it helps.

src/aiortc/rtcrtpsender.py
jlaine: This is unrelated to the problem at hand, let's keep this out of this commit.

jlaine:
I really appreciate the effort you're putting into these investigations, but can we please just solve one problem at a time? I would like to fix the bugs you're running into, and keep the "don't make RTCPeerConnection close tracks" well separated, as this is a user-facing change which will need to be flagged with a big red warning.

src/aiortc/contrib/media.py
jlaine: OK, I don't think this is actually a threading problem, but you are right: flow can block a few lines higher on the `frame = await self._queue.get()` line since it's an async statement, so `self._player` may have been cleared in the meantime. This would result in an exception in the caller (`RTCRtpSender`) which would therefore never properly reach the point where it sets the `_rtp_exited` event.

src/aiortc/contrib/media.py
```diff
@@ -223,6 +224,10 @@ def __init__(self, file, format=None, options={}):
         self.__thread: Optional[threading.Thread] = None
         self.__thread_quit: Optional[threading.Event] = None
 
+        # check whether we need to throttle playback
+        container_format = set(self.__container.format.name.split(","))
+        self._throttle_playback = not container_format.intersection(REAL_TIME_FORMATS)
```
jlaine: Are you sure we need this change? There is no way the `recv` methods of the tracks can be called before the constructor completes, so I don't see what difference it makes setting this earlier

src/aiortc/contrib/media.py
jlaine: I think two changes would make sense here:

- check `self._player is not None` as you did
- make `RTCRtpSender._run_rtp` more robust, by moving its shutdown code (including the critical `self.__rtp_exited.set()` in a `finally` clause of the try / catch

jlaine: @jmillan in a5764dc163351394f61108cd39f7351608fd90bc I made a change which ensures `RTCRtpSender._run_rtp` now catches and logs any unexpected exceptions and exits cleanly. Could you try running your tests without any of your fixes, you should now see an `AttributeError`?

ibc: I'm not running those tests but indeed, after `git pull` in aiortc "master", I see such an error:

```
sender(video) - RTP worker encountered an exception
  File "/usr/local/lib/python3.7/site-packages/aiortc/rtcrtpsender.py", line 295, in _run_rtp
    payloads, timestamp = await self._next_encoded_frame(codec)
  File "/usr/local/lib/python3.7/site-packages/aiortc/rtcrtpsender.py", line 248, in _next_encoded_frame
    frame = await self.__track.recv()
  File "/usr/local/lib/python3.7/site-packages/aiortc/contrib/media.py", line 174, in recv
    if self._player._throttle_playback and frame_time is not None:
AttributeError: 'NoneType' object has no attribute '_throttle_playback'
```

ibc: I've done some tests and I may say that, after that commit, I am not able to reproduce the "zombie process" issue anymore. This is, no matter I stop the mic/webcam tracks before/after closing the `pcs`, there are no more Python zombie processes. This is good.

jlaine: OK that is excellent news. I've had to back the commit out of master as CI was failing and I've moved the code to PR #293 to examine what's going on. Sorry for the merge conflict you're about to experience..

ibc: n.p. BTW I'm testing `transceiver.sender.replaceTrack()` and I get an error:

```
sender(video) - RTP worker encountered an exception
  File "/usr/local/lib/python3.7/site-packages/aiortc/rtcrtpsender.py", line 295, in _run_rtp
    payloads, timestamp = await self._next_encoded_frame(codec)
  File "/usr/local/lib/python3.7/site-packages/aiortc/rtcrtpsender.py", line 248, in _next_encoded_frame
    frame = await self.__track.recv()
  File "/usr/local/lib/python3.7/site-packages/aiortc/contrib/media.py", line 174, in recv
    if self._player._throttle_playback and frame_time is not None:
AttributeError: 'NoneType' object has no attribute '_throttle_playback'
```

jlaine: This is very much related, it's the same problem @jmillan pointed to, and which is addressed by PR #294. The difference is that we now have a visible error rather than a silent deadlock.

jlaine:
I'm going to close this PR, as PR #293 fixes the deadlock (and gives us a usable logging out) and PR #294 is going to fix the root cause of the deadlock.

Once both PRs are merged I will release aiortc 0.9.26.

jmillan:
:+1: :-)

src/aiortc/contrib/media.py
ibc: Sorry if unrelated. Randomly, `replaceTrack()` does not fail (no error log), however no RTP is sent anymore in the corresponding transceiver. This is probably related to the stuff you are already fixing :)

ibc: Well, sorry for the noise. Actually, when `replaceTrack()` works (and I've remove the `track.stop()` in `rtcrtpsender.py`), it keeps sending RTP. Confirmed. However the remote (Chrome) cannot render those RTP during ~10 seconds and gets frozen video. Asking for a key frame via RTCP PLI does not make it work. After ~10 seconds, the receiver can render the video again.

PR #294, opened 2020-02-24
[media player] ensure self._player is not None before using it

jlaine:
self._player may have been cleared by a call to track.stop() while we
were waiting for a frame, so check it is not None before using it.

Fixes: #283

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -171,7 +171,11 @@ class PlayerStreamTrack(MediaStreamTrack):
         frame_time = frame.time
 
         # control playback rate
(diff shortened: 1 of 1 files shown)


jmillan:
```bash
git remote add jlaine https://github.com/jlaine/aiortc.git
git fetch jlaine
git checkout jlaine/media-player-exception
git cherry-pick 2b4fce4b58aa3c56a05a112ab4d0634e0c45a3a6 // These are my tests
python3 -m unittest tests.test_rtcpeerconnection.RTCPeerConnectionTest
```

Result:
```txt
Ran 60 tests in 49.072s

OK
```

Issue #256, opened 2020-01-24
RTCDataChannel stream id never reuse


stalkerg:
In the transport code, you never reset number to 0 if it reaches the max. 
https://github.com/aiortc/aiortc/blob/master/src/aiortc/rtcsctptransport.py#L1620
```python
        if stream_id is None:
                while self._data_channel_id in self._data_channels:
                    self._data_channel_id += 2
                stream_id = self._data_channel_id
                self._data_channels[stream_id] = channel
                self._data_channel_id += 2
                channel._setId(stream_id)
```

it can be ok for a custom client but on the browser, we have a very strict limitation:
stream_id < 1024 on Chrome and stream_id < 256 on Firefox. 
On the Chrome side, you can check MaxChannels param. 
As you can see, if I need often close and create DataChannel I very quickly reach this limitation. 

Also, some details here https://github.com/w3c/webrtc-pc/issues/2158 

I suppose should be a way to setup max_stream_id and try reused freed ids.

jlaine:
I'm unfamiliar with the concept of recycling stream ids, are you sure that's a thing?

stalkerg:
Yes, I am sure and make a test application - I even find this limitation in Chrome sources. 
Please check my link. 

It looks like browsers preallocate structure or array for all possible channels. 

Currently, to avoid this situation I start to make all new channels on browser-side and wrote a simple manager.

lgrahl:
@jlaine You can reuse stream IDs once a stream has been reset from both sides.

jlaine:
OK @stalkerg feel free to propose a PR

jlaine:
@lgrahl the PR above does recycle stream IDs. However it still doesn't ensure we don't exceed the maximum allowed streams. Which limit should I be comparing to as there is a notion of inbound / outbound streams in SCTP?

lgrahl:
Against the minimum of the negotiated amount of incoming and outgoing SCTP streams. See steps 3 and 4 in https://w3c.github.io/webrtc-pc/#sctp-transport-connected.

jlaine:
Thanks @lgrahl !

@stalkerg would you mind giving the PR a try and report back whether it solves your problem?

NOTE: I am well aware there is currently no check that we have not exceeded the maximum number of allowed streams, but at least your stream IDs should get recycled.

PR #307, opened 2020-02-26
[data channels] look for the lowest available stream id

jlaine:
Fixes: #256

### DIFF
--- a/src/aiortc/rtcsctptransport.py
+++ b/src/aiortc/rtcsctptransport.py
@@ -680,6 +680,15 @@ class RTCSctpTransport(AsyncIOEventEmitter):
         return self.transport.transport.role != "controlling"
 
     @property
+    def maxChannels(self) -> Optional[int]:
+        """
+        The maximum number of :class:`RTCDataChannel`s that can be used simultaneously.
+        """
+        if self._inbound_streams_count:
+            return min(self._inbound_streams_count, self._outbound_streams_count)
+        return None
+
+    @property
     def port(self) -> int:
         """
         The local SCTP port number used for data channels.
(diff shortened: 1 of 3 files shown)


stalkerg:
Thanks, it's should be enough. I have a very similar code in the browser code now. 
Anyway, you can't check the maximum amount of channels from the "server". 
It will be great only add in documentation somewhere some warnings about current browsers limitations because behaviour then you try to open more channels from the server very odd.  (without any exceptions and etc)

lgrahl:
> Anyway, you can't check the maximum amount of channels from the "server".

Sure you can. It's the minimum of the negotiated amount of incoming and outgoing streams. See `maxChannels` in the WebRTC spec.

stalkerg:
@lgrahl I mostly talking about aiortc. Also even on browser-side this param announcing in Chrome only. 
But if this param exist and browsers are using it, I suppose it's great.

jlaine:
> Sure you can. It's the minimum of the negotiated amount of incoming and outgoing streams. See `maxChannels` in the WebRTC spec. It should be easy to expose that field. slightly_smiling_face

What I'm not finding in the specs is the expected behaviour if additional streams are negotiated using a RE-CONFIG chunk : does maxChannels get raised to reflect the additional streams?

lgrahl:
> What I'm not finding in the specs is the expected behaviour if additional streams are negotiated using a RE-CONFIG chunk : does maxChannels get raised to reflect the additional streams?

While renegotiation is part of this extension spec of SCTP, it is not mandatory for WebRTC data channels. But, sure, the answer would be yes.

PR #314, opened 2020-03-02
Don't use f-strings to format log messages

uriyyo:
Replace f-strings from `__debug_log` function call.
This pull request related to #262.

### DIFF
--- a/src/aiortc/rtcdatachannel.py
+++ b/src/aiortc/rtcdatachannel.py
@@ -199,7 +199,7 @@ class RTCDataChannel(AsyncIOEventEmitter):
 
     def _setReadyState(self, state: str) -> None:
         if state != self.__readyState:
-            self.__log_debug(f"- {self.__readyState} -> {state}")
+            self.__log_debug("- %s -> %s", self.__readyState, state)
             self.__readyState = state
 
             if state == "open":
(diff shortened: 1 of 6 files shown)


jlaine:
Thanks for jumping on this so quickly! However I don't know whether we need to go as far as rolling everything back.

To be clear, not all log messages happen very often. For instance the log messages which occur once per session, or under error conditions are probably fine. The more problematic ones are those which log every single RTP / RTCP packet, every SCTP chunk, or every SCTP timer.

I think we have several options:

- switch all debugging log statements back to the old format, as you did in this PR

- only do this for the performance sensitive ones

- entirely disable the performance sensitive log statements if we are not in debug mode

Any idea how other projects handle this?

uriyyo:
I think the best option, in this case, will be to switch all debugging log statements back to the old format.

In such case, `logging` module will format messages only in case when appropriated debug level set. 

The only disadvantage of this approach is that we must use old-style string formating for log messages.

What your point of view about that?

lgrahl:
> Any idea how other projects handle this?

Keeping formatters and arguments separate. :slightly_smiling_face: Agree with @uriyyo here. However, for really expensive operations, there's Python's `__debug__` built-in variable. See https://logbook.readthedocs.io/en/stable/performance.html#debug-only-logging

PR #347, opened 2020-04-29
Allow custom bind address in examples

benwilber:
Allow a custom bind address in example scripts.

By default, the example scripts try to bind to `0.0.0.0` which triggers firewall prompts to allow on macOS.  This allows running the examples scripts on `127.0.0.1` by doing `--bind=127.0.0.1`.

### DIFF
--- a/examples/server/server.py
+++ b/examples/server/server.py
@@ -195,6 +195,7 @@ if __name__ == "__main__":
             host = args.bind
             port = args.port
     else:
+        host = "0.0.0.0"
         port = args.port
 
     if args.verbose:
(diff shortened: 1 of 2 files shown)


jlaine:
Hi!

I'm not opposed to having an option which controls which address you bind to, but I'd like to avoid confusion by having two ways to set the port. Why not just add a `--host` option which remains orthogonal to to `--port`?

benwilber:
Yes good call.  I've updated so it take --host argument instead

jlaine:
Thanks for the PR!

PR #375, opened 2020-06-04
Support for custom attributes in SDP

thedilletante:
Hi.

First of all, thanks for the great project. I use a small part of it in some personal experimental projects and as testing tool and actually fall in love with it. Imagine what will happen when I get used to the majority of API...

I found that for my needs the API of `aiortc.sdp.SessionDescription` lacks support of custom (non-standard) attributes, that could be used to implement additional features above SDP protocol.

I decided not to fire issue but create a PR and here it is.

The API for fetching custom attributes is simple: 
```
sdp = SessionDescription.parse("...")
sdp.attributes
```
Not to break current behavior I didn't implement default assembling of custom attributes. Instead I added `assemble_custom_attributes` boolean field into `SessionDescription` and `MediaDescription`. So user could choose to assemble them just updating a field of any SDP components. Also, `SessionDescription.set_assemble_custom_attributes()` method is added to update the field in `SessionDescription` and all of the related `MediaDescription` lines.

I would be glad to get feedback if suggested changes is acceptable for the project and is useful for somebody else.
Thanks.

### DIFF
--- a/src/aiortc/sdp.py
+++ b/src/aiortc/sdp.py
@@ -191,6 +191,10 @@ class MediaDescription:
         self.ice_candidates_complete = False
         self.ice_options: Optional[str] = None
 
+        # Other attributes
+        self.attributes: List[Tuple[str, str]] = []
+        self.assemble_custom_attributes = False
+
     def __str__(self) -> str:
         lines = []
         lines.append(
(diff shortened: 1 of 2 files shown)


jlaine:
Hello! Thanks for taking the time to put together a PR. I'm having trouble understanding what the usecase for this is: the SDP parsing is internal code which has no public API, so how does supporting custom attributes benefit `aiortc` users?

thedilletante:
Hello, @jlaine !
Thank you for the quick reply.
I agree that SDP manipulation is not the main purpose of the project.
But as I said currently I use only limited subset of the `aiortc` to communicate with webrtc signalling server (without media management). So when I was looking for existing webrtc SDP parser/assembler API for python I didn't find the implementation better than in `aiortc`.
Also, it is more easier to do SDP transformations in the middle of the path from `RTCPeerConnection.localDescription` to signalling server (which is the case when signalling server requires some custom attributes in SDP).

PR #392, opened 2020-07-14
WIP: Computer vison streaming demo

kodonnell:
I won't write too much here as there's a bunch in the readme.md and you should be able run the demo code easily to see what it's talking about. But I think it's pretty cool (naturally) and will be useful to a bunch of people.

Status:
- Code is working aside from some heisenbugs (as documented in the readme). I wouldn't mind someone who knows about webrtc/aiortc to run their eyes over it to see if I'm doing it the "right" way, or if there are more optimal ways. Possibly @jlaine (partly as I'm also hoping it'll tickle your fancy).
- Docs etc. need to be updated with some screenshots etc. I've written them from my perspective (e.g. "I'm a novice and don't trust me") which we might want updated if it's going in here (e.g. "less novice people have reviewed this and OK'd it").


examples/computer-vision-server-side-streaming/readme.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Use the WebRTC stats which, as you've seen, has the number of frames received. This is effectively the frame number, right? My main concern here is getting it in sync (i.e. knowing exactly which frame is the first that gets displayed and increments the count to one), and how poor connection affects things (e.g. if some frames are dropped, and the count not incremented, then everything will be out). And, I guess, how it might vary from browser-to-browser, etc.
+2. Do some steganography! Basically, encode the frame number in the pixels of the image, and then decode it on the frontend. The challenges on this front:
```
inactivist, line 85: Is it possible to interleave metadata or other non-image "chunk" data in the encoded video stream? (I'm rusty on this, but IIRC it should be possible.)

kodonnell: Great question! In my experience, I haven't seen anything that works easily. The trick is, like you say, having access to it when you write *and* being able to get it out when you render it in the browser. I think I've seen `side_data` before but never looked in detail, so it definitely could work. (Likewise I've seen some things done with the subtitle.) Anyway - I would be *very* happy if you could figure it out - that'd be awesome!

kodonnell: Oooh, a colleague just found this https://github.com/w3c/webrtc-insertable-streams/blob/master/explainer.md Which might be the trick for this last bit. I'll update now.

jlaine:
Do you plan to complete this PR or should I close it for now?

kodonnell:
Hmmm I think it was actually finished IIRC. I'm probably not going to put any more time into it - up to you whether it's useful as-is for an example (even if not polished).

rprata:
@jlaine can we close this PR?

PR #394, opened 2020-07-15
fix for chrome NALU maxium number limit

buendiya:
### Environment
Chrome Version: 71.0.3578.98
Number of CPU Cores: 12
test video file: https://drive.google.com/file/d/1bernB0ta5176CW0mMn9QkQ4fdV9u_iSB/view

### Problem
When chrome receive video from aiortc, the frame rate is 1. Start chrome from command line with --verbose, it will output this warning:

> Received packet containing more than 10 NAL units. Will not keep track sps and pps ids for all of them.

### Reason
When use libx264 as encoder, it outputs too many small NAL units. Then _packetize_stap_a() function will aggregate these small NAL units. If the count of NAL units it aggregates is more than 9, the problem is triggered.

#### Solution 1: limit the maxium number of  NAL unit it can aggregate
As the code shows.

#### Solution 2: limit the number of NAL unit libx264 can output

[X264 doc](http://www.chaneru.com/Roku/HLS/X264_Settings.htm)

 The number of NAL unit libx264 output is determined by these parameters

1. slices
2. sliced-threads
3. threads

We can set the sliced-threads to off and set slices to little than 9, or set threads to little than 9.

BTW. If the video content is still, the count of NAL unit will be larger.

### DIFF
--- a/scripts/fetch-vendor
+++ b/scripts/fetch-vendor
@@ -28,10 +28,6 @@ parser.add_argument("--config-file", default=__file__ + ".json")
 args = parser.parse_args()
 logging.basicConfig(level=logging.INFO)
 
-# work around buggy 'tar' executable on Windows
-if sys.platform == "win32":
-    os.environ["PATH"] = "C:\\Program Files\\Git\\usr\\bin;" + os.environ["PATH"]
-
 # read config file
 with open(args.config_file, "r") as fp:
     config = json.load(fp)
--- a/scripts/fetch-vendor.json
+++ b/scripts/fetch-vendor.json
@@ -1,3 +1,3 @@
 {
-    "urls": ["https://github.com/aiortc/aiortc-codecs/releases/download/1.1/codecs-{platform}.tar.gz"]
+    "urls": ["https://github.com/aiortc/aiortc-codecs/releases/download/1.3/codecs-{platform}.tar.gz"]
 }


jlaine:
I'm not sure I understand where the magic value "9" comes from, could you clarify this please?

buendiya:
> I'm not sure I understand where the magic value "9" comes from, could you clarify this please?

According to [https://github.com/JumpingYang001/webrtc/blob/ee01a839d2ed0a3b2a4e58019ea9a37a614e0d8a/modules/video_coding/codecs/h264/include/h264_globals.h#L64](url)
and
[https://github.com/JumpingYang001/webrtc/blob/264bee8bab9ba0a50c360ae34c9ebf801c17078b/modules/rtp_rtcp/source/rtp_format_h264.cc#L623](url)
and
[https://github.com/JumpingYang001/webrtc/blob/264bee8bab9ba0a50c360ae34c9ebf801c17078b/modules/rtp_rtcp/source/rtp_format_h264.cc#L685](url)
It seems that when the count of NAL units is equal 10, it will output the warning and discard the NAL units.
So the the count of NAL units must be smaller than 10.

buendiya:
> I'm not sure I understand where the magic value "9" comes from, could you clarify this please?

According to https://github.com/JumpingYang001/webrtc/blob/ee01a839d2ed0a3b2a4e58019ea9a37a614e0d8a/modules/video_coding/codecs/h264/include/h264_globals.h#L64
and
https://github.com/JumpingYang001/webrtc/blob/264bee8bab9ba0a50c360ae34c9ebf801c17078b/modules/rtp_rtcp/source/rtp_format_h264.cc#L623

When the count of NAL units is more than 10, it will output the warning and discard the NAL units.
But if I set the counter to 10, it still outputs the warning. If set to 9, it works normally.
Sorry. I don't know why it can't be 10.

jlaine:
Could you please rebase your patch on top of the latest code to fix CI and add a unit test to capture this case?

buendiya:
> Could you please rebase your patch on top of the latest code to fix CI and add a unit test to capture this case?

I have added one test. Sorry. It takes so long.

jlaine:
Sorry I messed up somehow, I've created a fresh PR here with your changes: #426

PR #403, opened 2020-08-11
Fix disappearing start/stop button in examples/server

Raghav-B:
One line change, just made it so that the `Start` button on the html page reappears when clicking `Stop`. This allows for setting up a new RTCPeerConnection without refreshing the entire page.

### DIFF
--- a/examples/server/client.js
+++ b/examples/server/client.js
@@ -207,6 +207,8 @@ function stop() {
     setTimeout(function() {
         pc.close();
     }, 500);
+
+    document.getElementById('start').style.display = 'inline-block';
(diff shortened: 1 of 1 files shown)


jlaine:
The button "disappears" for a reason: the code was not written to allow it to run multiple times.

Raghav-B:
@jlaine My bad, I can try changing the code so that it works properly multiple times? Or I can just close this pull request.

jlaine:
You're welcome to try adapting the code, but let's make sure it remains easy to read as that's the purpose of the examples :)

PR #434, opened 2020-11-17
Using preferred codecs list to configure local RTP stream. Adding possibility to use an encoded stream track without re-encoding

jpiat:
this commit allows setting up the preferred codecs to be used for the local RTP stream. By default aiortc uses VP8 but on some platform one may enforce the use of H264 to minimize CPU load (on embedded platforms such as the raspberry pi).


jpiat:
can anyone at aiortc give a feedback on this pull request or let me know what i should do to get this pull request merged in ?

jan-bogaerts:
Hi, I like the approach that you are taking. I was trying to do something similar for an IP camera that produces a H264 stream, but I got stuck on the class that is similar to your PiTrack. I use this class to read packets from the av container. Something goes wrong though, cause there is no video on the other side. I have written an issue about it here:
https://github.com/aiortc/aiortc/issues/448
Could you help me out with this perhaps?

jan-bogaerts:
Found the problem: you apparently changed a little bit in the _split_bitstream function. Once I went back to the original function as in the h264.py module, video started appearing.
On to the next problem, stopIteration. probably need a buffer somewhere. Anyway, sorry to bother you.

jpiat:
Thanks for giving it a try ! I actually want to do exactly what you did,
using a RTSP IP camera as an input using a gstreamer pipeline. I'm going to
push a new example soon for this.

I modified the _split_bitsream function for a pi encoder output because i
know the bistream structure in each buffer and it allowed to minimize a but
the processing.  To support any bistream from any camera the initial
approach might work better.

Can you describe the stopIterarion problem ? How can I help ?

Le mar. 12 janv. 2021 à 19:46, Jan Bogaerts <notifications@github.com> a
écrit :

jan-bogaerts:
Hi, Thanks.
The stopIteration problem, I believe, is due to the camera that hangs for a short while after initial connection. Once the application gets over that hurdle, the connection appears stable. When using ffplay, the connection was definitely stable, but there was also this stutter at the beginning. 
When I tried with the original aiortc 'server' example, which does the heavy conversion, there would often be the same errors during initial startup, but if the app did manage to get through that initial startup, it would be ok (read-buffers would fill up anyway cause conversion could not follow).
In the code that I was testing yesterday evening, I was doing this (in my version of EncodedStreamTrack:
```
async def recv_encoded(self, keyframe=False) -> List[bytes]:
        while True:
            packet = next(self.demuxer)
            if not packet.dts is None: break
        timestamp = convert_timebase(packet.pts, packet.time_base, VIDEO_TIME_BASE)
        return self._packetize(self._split_bitstream(packet.to_bytes())), timestamp
```
So I figure it has something to do with this `packet = next()` line or the caller of this function which also appears to be wrapped in a 'next()' call.
I was thinking of doing something similar along the lines that you did with this nal_queue and separate thread for reading the packets. What do you think?
Also, I did a small thing different in my version of picamera.py to set up preference for 'h264'. I did it like this on the server side:

```
    capabilities = RTCRtpSender.getCapabilities("video")
    preferences = list(filter(lambda x: x.name == "H264", capabilities.codecs))
    #preferences += list(filter(lambda x: x.name == "VP8", capabilities.codecs))
    #preferences += list(filter(lambda x: x.name == "rtx", capabilities.codecs))
    transceiver = pc.getTransceivers()[0]
    transceiver.setCodecPreferences(preferences)
```
I found that bit of code in one of the other issues over here (the code comes from the tests included with aiortc)
In the end, I simply removed vp8 and rtx. Kept working. On the other side (client), I found something you mentioned (that it also had to be changed) and currently did something hacky like this:
```
offer.sdp = offer.sdp.replace("UDP/TLS/RTP/SAVPF 96 97 98 99 100 101 122 102 121 127 120 125 107 108 109 124 119 123 118 114 115 116", "UDP/TLS/RTP/SAVPF 102 125 108 99 100 101 122 96 121 127 120 97 107 98 109 124 119 123 118 114 115 116");
```
before calling `pc.setLocalDescription(offer)`

not certain yet if all of this is needed. Just letting you know, cause I think you did it differently?

jan-bogaerts:
FYI, this managed to get the video feed stable:

```
import av
import threading
from queue import Queue

class StreamPlayer (threading.Thread):
    def __init__(self, inputUrl):
        threading.Thread.__init__(self)
        self.inputUrl = inputUrl
        self.isRunning = False                       # flag to indicate that the thread should stop
        self.packets = Queue(40)
        self.container = None                       # the pyav container

    def run(self):
        """
        start the thread until a stop is requested.
        :return:
        """
        print("starting player thread")
        self.isRunning = True
        if self.inputUrl.startswith('rtsp'):
            options = {'rtsp_transport': 'tcp'}
        else:
            options = {}
        source = av.open(self.inputUrl, mode="r", metadata_encoding='utf-8', options=options)
        self.container = source                                                     # for outside ref

        while self.isRunning:
            for i, packet in enumerate(source.demux()):
                try:
                    if packet.dts is None:
                        continue

                    if packet.stream.type == 'video':
                        if not self.packets.full():
                            self.packets.put(packet)
                    if self.isRunning == False:
                        break

                except InterruptedError:
                    self.isRunning = False
                    break
        source.close()
```

jpiat:
Great ! i just added an example to read from a h264 webcam but this example
could also be used to read a rtsp stream using the rtsp src plugin in
gstreamer

Le mer. 13 janv. 2021 à 13:49, Jan Bogaerts <notifications@github.com> a
écrit :

jan-bogaerts:
If I understand it correctly, you work with the gstreamer library instead of pyAv-ffmpeg? just curious, but why exactly? Is there an advantage somewhere?

jpiat:
the main advantage that i see is that a number of single board computers
have a gstreamer plugin to support the built-in h264 encoder  (rk3399
boards, NVidia Jetson and the raspberry pi ...) but not all have a ffmpeg
encoder. If we can have pyAv and Gstreamer that's even better !

Le mer. 13 janv. 2021 à 14:24, Jan Bogaerts <notifications@github.com> a
écrit :

jlaine:
This commit is so massive, I can't even tell what it's about.

The title mentions being able to configure the preferred codecs, but there's already an API for that?

Anything related to "sending out pre-encoded media" should be in its own PR.

jpiat:
There is no API (at least not documented) to configure the preferred codec
in the offer. By default the offer always contains all the available
codecs, what i did is that the preferred codecs list is also used to filter
the codecs in the offer. This allow to disable VP8 in the offer when the
codec is not available or does not perform well on the sender architecture.

I apologize for the messy PR but it started on the 17th of november (more
than two month ago) and i could not wait for it to be allowed to move
forward. If you look closely there is not much change in the aiortc code
itself (addition of an EncodedTrack type, and preferred codec list
filtering the offer), most of the change is additional examples to use
encoded source.

Le ven. 29 janv. 2021 à 13:01, Jeremy Lainé <notifications@github.com> a
écrit :

jlaine:
OK, feel free to come back with an alternative PR when you are willing to put in the effort.

Configuring codec preferences works in the standard way using `setCodecPreferences`:

https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences
https://aiortc.readthedocs.io/en/latest/api.html#aiortc.RTCRtpTransceiver.setCodecPreferences

jpiat:
Hi,

i will. Regarding setCodecPreferences, i understand that there is a
standard way to set codec preference, but in your implementation there is
no way for the sender to reduce the list of supported encoding profiles. In
your implementation the offer always contains VP8 but in some cases it
would be helpful to remove it from the offer if your platform does not
support VP8 encoding in hardware. If its not through setCodecPreferences,
it should be possible through some other way.

Le mer. 17 févr. 2021 à 13:27, Jeremy Lainé <notifications@github.com> a
écrit :

jan-bogaerts:
>  Regarding setCodecPreferences, i understand that there is a
> standard way to set codec preference, but in your implementation there is
> no way for the sender to reduce the list of supported encoding profiles. In
> your implementation the offer always contains VP8 but in some cases it
> would be helpful to remove it from the offer if your platform does not
> support VP8 encoding in hardware. If its not through setCodecPreferences,
> it should be possible through some other way.
I agree with this. I was facing similar issues. Right now I hacked my way through it, but it would be convenient if the library supported this.

jlaine:
> In your implementation the offer always contains VP8 but in some cases it would be helpful to remove it from the offer if your platform does not support VP8 encoding in hardware. If its not through setCodecPreferences, it should be possible through some other way.

Using setCodecPreferences allows you to reorder codecs or even completely remove a codec from the offer.

jpiat:
no it does not ...  look into the method of rtcpeerconnection __localRtp here
the codecs list is unfiltered.

Le mer. 17 févr. 2021 à 17:12, Jeremy Lainé <notifications@github.com> a
écrit :

jlaine:
@jpiat give me a little credit, I wrote the code.

https://github.com/aiortc/aiortc/blob/ecb98815ceb00457ab4480cc96dc87197ca30bd3/src/aiortc/rtcpeerconnection.py#L595

jpiat:
I give you credit, but i tried setting codec preferences to h264 only but
the sent SDP still had VP8 listed. Give me a little credit, i tried.

Le mer. 17 févr. 2021 à 18:26, Jeremy Lainé <notifications@github.com> a
écrit :

jlaine:
I've just added an explicit test to the test suite to remove VP8 from the offer using `setCodecPreferences` in d828065626772f4058ee659cdb3708b379dfb673.

It works as expected.

jpiat:
If you define the preference and use it on the transceiver :

pi_capability =RTCRtpCodecCapability( mimeType="video/H264",
clockRate=90000, channels=None, parameters=codec_parameters)
preferences = [pi_capability]


# later
t.setCodecPreferences(preferences)

VP8 will still be part of the offer (at least it was 4 month ago)



Le mer. 17 févr. 2021 à 18:32, Jonathan Piat <piat.jonathan@gmail.com> a
écrit :

jpiat:
I just did a pull of latest aiortc, and modified webcam.py example with the following: 

```
codec_parameters = OrderedDict(
    [
        ("packetization-mode", "1"),
        ("level-asymmetry-allowed", "1"),
        ("profile-level-id", "42001f"),
    ]
)

h264_capability = RTCRtpCodecCapability(
    mimeType="video/H264", clockRate=90000, channels=None, parameters=codec_parameters
)

preferences = [h264_capability]
```


later to be applied to the video transceiver 

```
for t in pc.getTransceivers():
        if t.kind == "audio" and player.audio:
            pc.addTrack(player.audio)
        elif t.kind == "video" and player.video:
            pc.addTrack(player.video)
            t.setCodecPreferences(preferences)
```

and here are the codecs in the offer

```
a=rtpmap:96 VP8/90000
a=rtcp-fb:96 nack
a=rtcp-fb:96 nack pli
a=rtcp-fb:96 goog-remb
a=rtpmap:97 rtx/90000
a=fmtp:97 apt=96
a=rtpmap:102 H264/90000
a=rtcp-fb:102 nack
a=rtcp-fb:102 nack pli
a=rtcp-fb:102 goog-remb
a=fmtp:102 packetization-mode=1;level-asymmetry-allowed=1;profile-level-id=42001f
a=rtpmap:121 rtx/90000
a=fmtp:121 apt=102
a=rtpmap:125 H264/90000
a=rtcp-fb:125 nack
a=rtcp-fb:125 nack pli
a=rtcp-fb:125 goog-remb
a=fmtp:125 packetization-mode=1;level-asymmetry-allowed=1;profile-level-id=42e01f
a=rtpmap:107 rtx/90000
a=fmtp:107 apt=125
```

using chrome://webrtc-internals , the codec selected is indeed VP8.

PR #445, opened 2021-01-09
fix Invalid element type (room should be a string)

JonathanXu1:
This fixes a minor bug in the Janus example. The default room (1234) is not a string, which causes an error response from Janus. This PR casts args.room to a string to fix this issue.

image

### DIFF
--- a/examples/janus/janus.py
+++ b/examples/janus/janus.py
@@ -246,7 +246,7 @@ if __name__ == "__main__":
     loop = asyncio.get_event_loop()
     try:
         loop.run_until_complete(
-            run(player=player, recorder=recorder, room=args.room, session=session)
+            run(player=player, recorder=recorder, room=str(args.room), session=session)
         )
     except KeyboardInterrupt:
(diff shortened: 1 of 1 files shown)


ruddell:
Do you have `string_ids` set to true in your [videoroom config file](https://github.com/meetecho/janus-gateway/blob/be588d8c1751d2ca8c5eac372ac2590ac134d907/conf/janus.plugin.videoroom.jcfg.sample#L54-L56)?  By default, the Janus Room ID should be a number:
>
	# By default, integers are used as a unique ID for both rooms and participants.
	# In case you want to use strings instead (e.g., a UUID), set string_ids to true.
	#string_ids = true

The [Janus JS demo](https://github.com/meetecho/janus-gateway/blob/ba0b83bbc93b070c3ff47a46eede4cb7d61c59d7/html/videoroomtest.js#L55-L57) uses a number instead of a string as well.

Did not test this myself, just going off the documentation.

Issue #335, opened 2020-04-14
Recording Janus Video room produces periodic aberations 


islandereleven:
Hi everyone,

I am encountering an issue whilst recording a Videoroom from a remote Janus server with the [examples/janus](https://github.com/aiortc/aiortc/blob/master/examples/janus/janus.py) code.

The initial recorded frames are consistently crisp but the recording eventually gets broken and eventually recovers (probably when a key frame is received) to eventually break again in a cyclical fashion. I have put sample images from a recording on [imgur](https://imgur.com/a/DqaO0ze). From my meagre understanding of WebRTC it would seem like it might have something to do with key frames getting lost and the PLI not kicking in. 

Has any one encountered similar issues?

This should be reproducible using the following command below adapting the room and janus server (the same phenomenon can be observed recording to mp4 as well).

```python3 janus.py --room janus_room --record-to %09drecording.png janus_server_url```

jlaine:
`aiortc` has the code to send a PLI (Picture Loss Indication), see `RTCRtpReceiver._send_rtcp_pli`

However I haven't had a chance to write the code which *triggers* sending it, so you are right, the image stays broken until the next keyframe. It's been a while since I dived into that part of the code, could you try poking around WebRTC specs and / or implementations to see under what condition we should be sending PLI?

islandereleven:
Ah ok, that's good to know. Thanks for the swift answer. 

I'll look up what needs to be done then.

jlaine:
I had forgotten but sending PLI is already tracked by issue #58 let's continue the discussion there.

Issue #196, opened 2019-07-18
example videostream-cli is Failing


Nishanth009:
Hi,
I ran videostream-cli example as per instructions provided in the readme file. The transmitted video from one peer reached the other(both python peers are in same machine). But after some time transmission stops and when i opened recorded video file, it was corrupted.
I also noticed this "using avstream.codec to pass codec parameters to muxers is deprecated" getting printed at terminal 
What would be the issue?Any thoughts?
Thanks in advance..

jlaine:
I really have nothing to go on here, no logs or any steps to reproduce..

The warning from ffmpeg is not the cause of your problem, this is just PyAV using some deprecated ffmpeg APIs.

Nishanth009:
There is no error which is getting printed. I could see the frames received by aiortc some of them are corrupted(first few frames are proper).I have attached samples of few received frames.What would be the reason for this.So i have added STUN server in peer connection, didn't solved the problem.I am using 4G network.
```
Also i have run cli.py example(videostream-cli) in different networks and it didn't work.The video frames are not received after offer and answer exchange
```
Reproducing these errors would be easy i guess since i used same example codes. I am attaching the links of samples videos i have tested 

https://sample-videos.com/video123/mp4/360/big_buck_bunny_360p_30mb.mp4

Any thoughts on these issues. Thanks in advance.


![video-178](https://user-images.githubusercontent.com/20472518/61621848-518de580-ac91-11e9-912e-79fb17e20d72.png)
![video-181](https://user-images.githubusercontent.com/20472518/61621849-52267c00-ac91-11e9-9c7d-c3c835f16b4d.png)
![video-182](https://user-images.githubusercontent.com/20472518/61621851-52267c00-ac91-11e9-9b33-4d878ab232b7.png)
![video-218](https://user-images.githubusercontent.com/20472518/61621853-52267c00-ac91-11e9-9bc4-8162964d56dc.png)
![video-232](https://user-images.githubusercontent.com/20472518/61621855-52bf1280-ac91-11e9-8a45-3dfef22aead2.png)

jlaine:
I'm still in the dark here, could you at the very least run the example in verbose mode (`-v`) and see if there are any interesting messages?

Glunky:
Hi! 
[...]

Issue #182, opened 2019-06-14
Video stream became 'broken' randomly


bymet:
Streaming can work fine for 10 seconds or 10 minutes, but then suddenly it becomes like that:
![image](https://user-images.githubusercontent.com/33255173/59494902-da498280-8e96-11e9-931b-fc87b55356cd.png)

Eventually it can fix itself, but sometime it does not. I have tried to stream from android device and browser and got same behavior. Any idea what can be wrong?

Frame processing looks like that:
```
async def recv(self):
        frame = await self.track.recv()

        if (self.transform == "nothing"):
            return frame

        try:
            networkWidth = self.detectionService.width
            networkHeight = self.detectionService.height
            
            widthScale = frame.width / networkWidth
            heightScale = frame.height / networkHeight

            frame_rgb = cv2.cvtColor(frame.to_ndarray(format="bgr24"), cv2.COLOR_BGR2RGB)
            frame_resized = cv2.resize(frame_rgb, (networkWidth, networkHeight), interpolation=cv2.INTER_LINEAR)

            darknet.copy_image_from_bytes(self.darknet_image, frame_resized.tobytes())
            self.lastDetection = self.detectionService.detect(self.darknet_image)
            self.detections.put(self.lastDetection)

            image = cvDrawBoxes(self.lastDetection, frame_rgb, widthScale, heightScale)
            new_frame = VideoFrame.from_ndarray(image, format="rgb24")
            new_frame.pts = frame.pts
            new_frame.time_base = frame.time_base

            return new_frame
        except:
            info = sys.exc_info()
            return frame;            
```

jlaine:
What is the screenshot I'm looking at? Something received by aiortc? Sent by aiortc?

The corruption you are showing looks like packet loss, so you'll want to figure out why you are losing packets.

Concerning the code pasted above, I'm not convinced it will run it realtime unless you use a low framerate + video size and / or offload the image processing to a separate thread or process. If you are doing 30fps you have at most 30ms to process each image, is this the case?

jlaine:
Hello?

andriy-tsvikla:
I am facing similar issue.   stream 320x240 4 fps

jlaine:
I'm going to close this issue for lack of response from the original submitter - I have nothing to go on here.

jlaine:
Should be fixed by 42f4e0f57b186aa15b0e731be37ca35afd780100

Issue #176, opened 2019-05-28
ValueError: packet is too long when trying to stream from v4l2 /dev/video*


hugobordigoni:
Hi,
I'm very grateful for your work on this package!

I'm working on videostream-cli example and I face several issues when streaming from a device (on ubuntu through v4l2 /dev/video).

The only change I made from videostream-cli is to use this media player: `MediaPlayer('/dev/video0', format='v4l2', options={'framerate': '30', 'video_size': '1280x720'})`

I launched the answer with `python cli.py answer --signaling tcp-socket --record-to output/%06d.jpg`. After several seconds or few minutes (depending on the execution), it stops reading frame. Using verbose mode, I have a message saying that dtls goes in CLOSED state. I added this in rtcdtlstransport.RTCDtlsTransport.__run (line 492):
```
except Exception as e:
    import traceback
    traceback.print_exc()
```

So now, I see that it was interrupted because this exception was raised:
```
  File "[...]/pylibsrtp/__init__.py", line 200, in protect_rtcp
    return self.__process(packet, lib.srtp_protect_rtcp, SRTP_MAX_TRAILER_LEN)
  File "[...]/pylibsrtp/__init__.py", line 224, in __process
    raise ValueError('packet is too long')
ValueError: packet is too long
```
I can send the full traceback if it's useful.

When I run the answer in verbose mode, I see a lot of packets_lost. And every time the exception is raised, the number of packet loss is around 5000.

Also, I don't know if it's related but when I take a look at the images, the first ones are good, but approximately when there is the first packet loss, the images get completely broken.  [Example image](https://ibb.co/KrpY0hd)

And the quality is worst at each frame, it's never going back to normal. Like if there was no key frame at all.

quito418:
Hi I am experiencing similar problems.

I came into 2 conclusions and trying to resolve the issue.

First, the image going wrong is because of packet loss in sending packets of 1 frame. (frame consists of lots of packets)
And if one frame goes wrong, the following frames goes wrong until the receiver gets a new key frame (you can check video tends to get better by changing the codec key frame interval to 1 second).

My conclusion now is 2 below

1. NACK has some issue. As RTP_HISTORY_SIZE in rtcrtpsender.py is small like 128, but receiver is sending nack for all missing packets where required histroy size is much bigger than 128, I think currently receiver is requesting retransmission of useless packets, and may forget retransmitting lost packet when frame size is bigger than 128 ( I did observed frame with 161 packets). 

[...]

PR #461, opened 2021-02-02
Jitterbuffer fix + basic send PLI implementation

Przem83:
Hi,
I've prepared a fix that may interest a lot of people having picture lost issues. I discovered that a big part of those issues might be related with a bad implementation of a jitter buffer that is used for collecting RTP packets before they are passed to the appropriate decoding function. You see, each of the RTP packets sent by a browser (sender) has a packet number that is used to properly group the packets into video frames on the receiver side. In theory those packet numbers should be unique, in practice however they are limited by the unsigned short int size which is 65535. Unfortunately the aiortc jitter buffer implementation did not handle the packet numbering return from 65535 to 0 correctly, causing the jitter buffer to be purged and the video signal to be lost or at least largely degraded. The situation repeated itself every 65535 received packets and was complicated by the fact that the aiortc does not support sending PLI (picture lost indicator) messages. Without the PLI messages being sent that video image could remain corrupted indefinitely as it is the video encoder decision if and when to send a new keyframe (in webRTC this usually happens on a rapid scene change or when it is explicitly requested through PLI).

This commit fixes the jitter buffer issue and adds a simple send PLI mechanism to the jitter buffer implementation.

### DIFF
--- a/src/aiortc/jitterbuffer.py
+++ b/src/aiortc/jitterbuffer.py
@@ -28,47 +28,38 @@ class JitterBuffer:
 
     def add(self, packet: RtpPacket) -> Optional[JitterFrame]:
         # print("[INFO][JitterBuffer] Received Seq:", packet.sequence_number)
-        delta = 0
         if self._origin is None:
             self._origin = packet.sequence_number
-        elif self._origin < MAX_MISORDER and packet.sequence_number > self.__max_number - MAX_MISORDER:
-            if self._origin >= (packet.sequence_number + MAX_MISORDER) % self.__max_number:
-                # print("[INFO][JitterBuffer] Received Seq1:", packet.sequence_number, "In Origin", self._origin,
-                #       'max_number of ', self.__max_number)
-                self.remove(self.capacity)
-                self._origin = packet.sequence_number
-                if self.sendPLI is not None:
-                    asyncio.ensure_future(self.sendPLI(packet.ssrc))
-            else:
-                return None
-        elif packet.sequence_number < MAX_MISORDER and self._origin > self.__max_number - MAX_MISORDER:
-            delta = (packet.sequence_number - self._origin) % self.__max_number
+            delta = 0
+            misorder = 0
         else:
-            if packet.sequence_number <= self._origin - MAX_MISORDER:
-                # print("[INFO][JitterBuffer] Received Seq2:", packet.sequence_number, "In Origin", self._origin,
-                #       'max_number of ', self.__max_number)
+            delta = (packet.sequence_number - self._origin) % self.__max_number
+            misorder = (self._origin - packet.sequence_number) % self.__max_number
+
+        if misorder < delta:
+            if misorder >= MAX_MISORDER:
+                # print("[INFO][JitterBuffer] Received Seq1:", packet.sequence_number, " Origin:", self._origin,
+                #       ' misorder:', self.__max_number)
                 self.remove(self.capacity)
+                # self.smart_remove(self.capacity, dumb_mode=True) # "remove" might still be a bit faster
                 self._origin = packet.sequence_number
+                delta = misorder = 0
                 if self.sendPLI is not None:
                     asyncio.ensure_future(self.sendPLI(packet.ssrc))
-            elif packet.sequence_number < self._origin:
-                return None
             else:
-                delta = (packet.sequence_number - self._origin) % self.__max_number
-        # print("[INFO][JitterBuffer] Received Seq3:", packet.sequence_number, "In Origin", self._origin,
-        #       'delta ', delta)
+                return None
 
-        if delta >= 2 * self.capacity:
-            # received packet is so far beyond capacity we cannot keep any
-            # previous packets, so reset the buffer
-            self.remove(self.capacity)
-            self._origin = packet.sequence_number
+        # print("[INFO][JitterBuffer] Received Seq2:", packet.sequence_number, " Origin:", self._origin,
+        #       ' delta:', delta)
+
+        if delta >= self.capacity:
+            # remove just enough frames to fit the received packets
+            excess = delta - self.capacity + 1
+            print("[WARNING][JitterBuffer] At least:", excess, "packets will be lost")
+            if self.smart_remove(excess):
+                self._origin = packet.sequence_number
             if self.sendPLI is not None:
                 asyncio.ensure_future(self.sendPLI(packet.ssrc))
-        elif delta >= self.capacity:
-            # remove just enough packets to fit the received packets
-            excess = delta - self.capacity + 1
-            self.remove(excess)
 
         pos = packet.sequence_number % self._capacity
         self._packets[pos] = packet
@@ -101,6 +92,7 @@ class JitterBuffer:
                 frames += 1
                 if frames >= self._prefetch:
                     self.remove(remove)
+                    # self.smart_remove(remove, dumb_mode=True) # "remove" might still be a bit faster
                     return frame
 
                 # start a new frame
@@ -117,3 +109,24 @@ class JitterBuffer:
             pos = self._origin % self._capacity
             self._packets[pos] = None
             self._origin = (self._origin + 1) % self.__max_number
+
+    def smart_remove(self, count: int, dumb_mode: bool = False) -> bool:
+        # smart_remove makes sure that all packages belonging to the same frame are removed
+        # it prevents sending corrupted frames to decoder
+        timestamp = None
+        for i in range(self._capacity):
+            pos = self._origin % self._capacity
+            packet = self._packets[pos]
+            if dumb_mode:
+                if i == count:
+                    break
+            elif packet is not None:
+                if i >= count and timestamp != packet.timestamp:
+                    break
+                timestamp = packet.timestamp
+            self._packets[pos] = None
+            self._origin = (self._origin + 1) % self.__max_number
+            if i == self._capacity - 1:
+                print("[Warning][JitterBuffer] JitterBuffer purged !!!")
+                return 1 
+        return 0
--- a/src/aiortc/jitterbuffer.py
+++ b/src/aiortc/jitterbuffer.py
@@ -1,7 +1,7 @@
-import asyncio
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 from .rtp import RtpPacket
+from .utils import uint16_add
 
 MAX_MISORDER = 100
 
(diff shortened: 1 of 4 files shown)


jlaine:
Could we please have a clean PR so there is a chance of actually merging it? If you have identified a bug related to packet number wrap-around, please submit a PR which addresses JUST that. The work on PLI seems interesting but incomplete, so this looks like a candidate for a separate PR.

Przem83:
What exactly do you mean by clean PR exactly? - do you want me to squash all commits into one or are you talking about separating the wrap-around fix  from the PLI part? From my perspective they both are a jitterbuffer fix as a properly implemented jitterbuffer should trigger a PLI request each time it is being purged (reset).

PS. The PLI implementation I proposed is simple but it is also complete (it addresses all the cases when a PLI should be sent). The only thing missing is the PLI re-transmission mechanism, but that can be a candidate for a separate PR as you said.

jlaine:
Yes if you could please split the wrap-around fix and the PLI part that would be perfect. Also make sure you remove any "print()" statements and commented lines of code please.

Przem83:
I've cleared the code as you asked. For the split of code i did create a separate branch with just the wrap-around fix, but before switching the branches in the PR I would like to try to convince you that pushing the PR as it is right now is actually a better idea - and here is why:
1) For static cameras (my use case), without the send PLI part, whenever there will be some kind of disturbance in the network the image will be permanently lost as there is no auto recovery mechanism in case the jitterbuffer is overflown. It is a big issue for anyone who want to use an aiortc connection for extended time periods as I do. 
2) In it's current form the PR should permanently fix the following issues: #26, #58, #182, #196, #335 and at least improve situation for #35, #28, #176
3) In my opinion making a separate PR for PLI implementation that is based on only two simple if statements is an overkill (almost no difference between the branch with and without PLI)  
4) Disabling the current send PLI implementation is as simple as removing a "sendPLI=self._send_rtcp_pli" parameter from JitterBuffer constructor (line 258 of rtcrtpreceiver.py)

I hope you do see the reasoning behind my point of view and that you'll agree on merging the fix in it's current form - if not then I'll of course switch the branches in the PR and we'll do it your way. I'm waiting for your final decision,
Best regards,
Przemek.

tests/test_jitterbuffer.py
```diff
@@ -114,14 +141,14 @@ def test_add_seq_too_high_discard_four(self):
         jbuffer.add(RtpPacket(sequence_number=1, timestamp=1234))
         self.assertEqual(jbuffer._origin, 0)
 
-        jbuffer.add(RtpPacket(sequence_number=2, timestamp=1234))
-        self.assertEqual(jbuffer._origin, 0)
+        #  jbuffer.add(RtpPacket(sequence_number=2, timestamp=1234))
```
jlaine, line 144: Please remove the code instead of commenting it.

tests/test_jitterbuffer.py
```diff
@@ -176,6 +203,26 @@ def test_remove(self):
         self.assertEqual(jbuffer._origin, 3)
         self.assertPackets(jbuffer, [None, None, None, 3])
 
+    def test_smart_remove(self):
+        jbuffer = JitterBuffer(capacity=4)
+
+        jbuffer.add(RtpPacket(sequence_number=0, timestamp=1234))
+        # jbuffer.add(RtpPacket(sequence_number=1, timestamp=1234))
```
jlaine, line 210: Please remove the commented line

src/aiortc/jitterbuffer.py
```diff
@@ -73,7 +83,7 @@ def _remove_frame(self, sequence_number: int) -> Optional[JitterFrame]:
                 # check we have prefetched enough
                 frames += 1
                 if frames >= self._prefetch:
-                    self.remove(remove)
+                    self.remove(remove)  # this might be a bit faster than smart_remove
```
jlaine, line 86: I don't understand the comment: should we use smart_remove() or not here?

src/aiortc/jitterbuffer.py
```diff
@@ -26,22 +29,29 @@ def capacity(self) -> int:
...
-            # received packet is so far beyond capacity we cannot keep any
-            # previous packets, so reset the buffer
-            self.remove(self.capacity)
-            self._origin = packet.sequence_number
-        elif delta >= self.capacity:
-            # remove just enough packets to fit the received packets
+            delta = 0
+            misorder = 0
+        else:
+            delta = (packet.sequence_number - self._origin) % self.__max_number
```
jlaine, line 35: Can we use uint16_add (defined in aiortc.utils) for consistency?

tests/test_jitterbuffer.py
```diff
@@ -4,6 +4,13 @@
 from aiortc.rtp import RtpPacket
 
 
+async def send_rtcp_pli(media_ssrc: int) -> None:
+    """
+    Send an RTCP packet to report picture loss.
+    """
+    print("[INFO][JitterBuffer] PLI sent to media_ssrc:", media_ssrc)
```
jlaine, line 11: This print() statement seems kind of pointless. Could we instead define a function in each test which stores a value which we can assert?

jlaine:
On the whole this PR looks very good. The one point I'm not very comfortable about is having synchronous code calling into asynchronous code. Is there any other way we could handle this?

src/aiortc/jitterbuffer.py
Przem83: Actually we can get rid of the remove() function entirely and replace it with smart_remove(). I haven't done so already as i didn't want to brake the backward compatibility and because smart_remove() checks the dumb_mode flag with each for loop tick (that's why it might be slightly slower). The smart_remove() run with dumb_mode flag set to true behaves exactly as the remove() function. I see 3 options here, we can:
1) leave it as it is
2) replace the remove() calls with smart_remove(dumb_mode=True) calls everywhere and get rid of remove() method
3) rename smart_remove() to remove() and replace the dumb_mode flag with smart_mode flag in the current implementation -  then we will have only remove(int) calls or remove(int, smart_mode=True) calls everywhere.

I leave the final decision to you as you now better how each of the changes can impact the rest of aiortc code.

src/aiortc/jitterbuffer.py
Przem83: I don't think so - I believe in python (-2 & 0xFFFF) = 2 while  (-2 % 65536) = 65534

tests/test_jitterbuffer.py
Przem83: Well you're right that it does nothing currently - I just thought it looks better then a pass statement. I'm not sure if I get what you want to achieve with the assert method though. For me this would be also pointless, as the async function will be run after the test finishes and we can not run a synchronous method with asyncio.ensure_future() (jitterbufer.py line 44 an 54).

tests/test_jitterbuffer.py
Przem83: No problem I'll remove it - I just thought it's easier to understand this way.

tests/test_jitterbuffer.py
Przem83: No problem I'll remove it - I just thought it's easier to understand what's going on this way.

Przem83:
OK - I've added some comments to your review and I'll wait for your answer before pushing the next commit. I see your point with running the asynchronous code from synchronous one - but I haven't found any easy solution to get around it. I did test the code very thoroughly though and confirmed that everything works as intended. If not doing this synchronous/asynchronous mix is really that important to you then I believe the best solution would be to rewrite the jitterbuffer code and make it asynchronous, as all the rtcp send methods I've found are asynchronous as well.

src/aiortc/jitterbuffer.py
jlaine: Nope they are identical:

```>>> -2 & 0xFFFF
65534
>>> -2 % 65536
65534```

It's good news too, otherwise all our sequence number calculations for what we send would be messed up..

File: src/aiortc/jitterbuffer.py [hunk 6]
```diff
@@ -89,4 +99,24 @@ def remove(self, count: int) -> None:
         for i in range(count):
             pos = self._origin % self._capacity
             self._packets[pos] = None
-            self._origin += 1
+            self._origin = (self._origin + 1) % self.__max_number
+
+    def smart_remove(self, count: int, dumb_mode: bool = False) -> bool:
+        # smart_remove makes sure that all packages belonging to the same frame are removed
```
jlaine, line 105: I missed this sorry, make this a docstring

tests/test_jitterbuffer.py
jlaine: My point is we are not asserting the function was actually called.

jlaine:
There is a very simple way of avoiding the call of avoiding the call to asynchronous methods: change the signature of `add()` to return `Tuple[Optional[JitterFrame], bool]`. The second member of the tuple would be "do we need to send a PLI?". That way the call to the async code is done by the caller, which in our case *is* an async method (so we won't use `ensure_future` at all):

https://github.com/aiortc/aiortc/blob/d56e14c57f88482dd5ac7d6018dd0f92a6ac12c3/src/aiortc/rtcrtpreceiver.py#L490

src/aiortc/jitterbuffer.py
```diff
@@ -89,4 +99,23 @@ def remove(self, count: int) -> None:
         for i in range(count):
             pos = self._origin % self._capacity
             self._packets[pos] = None
-            self._origin += 1
+            self._origin = (self._origin + 1) % self.__max_number
```
jlaine, line 102: `self._origin = uint16_add(self._origin, 1)`

src/aiortc/jitterbuffer.py
```diff
@@ -89,4 +99,23 @@ def remove(self, count: int) -> None:
...
+        timestamp = None
+        for i in range(self._capacity):
+            pos = self._origin % self._capacity
+            packet = self._packets[pos]
+            if packet is not None:
+                if i >= count and timestamp != packet.timestamp:
+                    break
+                timestamp = packet.timestamp
+            self._packets[pos] = None
+            self._origin = (self._origin + 1) % self.__max_number
```
jlaine, line 118: `self._origin = uint16_add(self._origin, 1)`

src/aiortc/jitterbuffer.py
```diff
@@ -18,35 +19,44 @@ def __init__(self, capacity: int, prefetch: int = 0) -> None:
         self._origin: Optional[int] = None
         self._packets: List[Optional[RtpPacket]] = [None for i in range(capacity)]
         self._prefetch = prefetch
+        self.__max_number = 65536
```
jlaine, line 22: Not needed with the suggested changes below

src/aiortc/jitterbuffer.py
```diff
@@ -18,35 +19,44 @@ def __init__(self, capacity: int, prefetch: int = 0) -> None:
...
+        else:
+            delta = uint16_add(packet.sequence_number, -self._origin)
+            misorder = uint16_add(self._origin, -packet.sequence_number)
+
+        if misorder < delta:
+            if misorder >= MAX_MISORDER:
+                self.remove(self.capacity)
+                self._origin = packet.sequence_number
+                delta = misorder = 0
+                if self._capacity >= 128:
```
jlaine, line 43: What's the 128 magic value? Could we have a named constant with a descriptive name?

src/aiortc/jitterbuffer.py
```diff
@@ -18,35 +19,44 @@ def __init__(self, capacity: int, prefetch: int = 0) -> None:
...
+                    pli_flag = True
+            else:
+                return pli_flag, None
+
+        if delta >= self.capacity:
+            # remove just enough frames to fit the received packets
             excess = delta - self.capacity + 1
-            self.remove(excess)
+            if self.smart_remove(excess):
```
jlaine, line 51: `smart_remove` seems to return a value.. which we ignore?

Przem83: We change it afterwards only when the buffer has been entirely purged (in such a case the smart_remove returns True). If we would delete the "self._origin = packet.sequence_number" from line 52, in case of situation where the buffer was purged we would and up with a self._origin value that points to a self._packets element that is None and this is a bad thing that would cause the buffer to be 100% filled no matter what (the remove_frame method would terminate on line 72 until the value would be overwritten).

Przem83:
I'll do another iteration with the code to address the suggestions from the newest review tomorrow. I think i should be able to push the next commit tomorrow before 15:00 (CET)

src/aiortc/jitterbuffer.py
Przem83: BTW we have the same situation with `remove` in line 40 and 41 of jitterbuffer.py. I've found that the previous code did not address the situation correctly when I've encountered some abnormal local network conditions, that caused a lot of packet drops.

src/aiortc/rtcrtpreceiver.py
```diff
@@ -487,7 +487,10 @@ async def _handle_rtp_packet(self, packet: RtpPacket, arrival_time_ms: int) -> N
             return
 
         # try to re-assemble encoded frame
-        encoded_frame = self.__jitter_buffer.add(packet)
+        pli_flag, encoded_frame = self.__jitter_buffer.add(packet)
+        # check if the PLI should be sent
+        if pli_flag:
+            asyncio.ensure_future(self._send_rtcp_pli(packet.ssrc))
```
jlaine, line 493: Don't use ensure_future, just await the call to send_rtcp_pli

src/aiortc/jitterbuffer.py
Przem83: This is the value set in rtcrtpreceiver.py (line 257) as the jitterbuffer size for video stream. I assume that this jitterbuffer size for video will rather increase in time, while the audio buffer should not reach 128 any time soon. I used the jitterbuffer size (capacity) to make sure that we are actually dealing with video before sending PLI.

Przem83: A viable alternative would be to add a "is_video" flag to the jitterbuffer `__init__` method, as currently there is no easy way to check if the jitterbuffer is used for video other then checking the `self._capacity` value.

jlaine: `is_video` sounds like a good idea!

jlaine:
Ok I think we've reached a good place and you have obviously taken the time to test this code. I'm going to merge it as-is, we can always have a follow-up PR. I'd actually love to see other PRs from you if you have time.

Thanks so much for bringing the coverage back to 100%!

jlaine:
Ah crap I forgot to squash the commits.

PR #465, opened 2021-02-10
[peer connection] add RTCPeerConnection.connectionState

jlaine:


Laityned:
Is there a reason to neglect the disconnected stage?

jlaine:
There is no support for the transient "disconnected" state in our RTCIceTransport implementation, because `aioice` does not emit any kind of "events" to report such a transient state.

Laityned:
On which source is this package based (that neglects the disconnected status)? This would help with creating a PR to include the disconnected stage.

Including the disconnected stage would be useful to find out when the other party has disconnected without explicit signaling.

jlaine:
aioice lives at https://github.com/aiortc/aioice/

I don't see what "connectionState" has to do with signaling? Also, `connectionState == "disconnected"` doesn't tell you that the remote party has gone away. It just means some STUN packets were not answered (which may a temporary situation). True loss of connection will cause a "failed" state.

Laityned:
For this the newest version on the main branch is used.

In the attachment a logfile is found whereby the browser is closed while the stream is running. 

As seen in the log file at 18:07:17 it is detected that the DTLS is shutdown. However this results never in a 'disconnected' or 'failed' as could be seen in the logs by the 'Peer connection state is ....' and the 'ICE connection state ...'. 

The ice connection as well as the peer connection are turning to the closed state when program is shutting down
[log.txt](https://github.com/aiortc/aiortc/files/5960336/log.txt)

jlaine:
Can you please open a new issue, including how to reproduce with one of the built-in repos?

Also, I see you have a "multiplexed source". I have been wanting to add something like this to `aiortc.contrib.media`, would you consider submitting a PR?

PR #467, opened 2021-02-10
[Develop] Settings option for dev in setup and update  gitignore

Laityned:
When installing editable option, enables option to install aiohttp and websockets to be able to run all tests local.

Update gitignore to exclude venvs

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,15 @@ install_requires = [
     "pylibsrtp>=0.5.6",
 ]
 
+extras_require = {
+    'dev': [
+        'aiohttp>=3.7.0',
+        'coverage>=5.0',
+        'numpy>=1.19.0',
+        'websockets>=8.0',
+    ]
+}
+
 if os.environ.get("READTHEDOCS") == "True":
     cffi_modules = []
     install_requires = list(filter(lambda x: not x.startswith("av"), install_requires))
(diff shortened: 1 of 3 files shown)


jlaine:
I'm not sure I see the point of these changes: 

- "venv" is an arbitrary name, I personally never call my virtualenvs like this
- why the "dev" extra_requires?

Laityned:
Maybe the gitignore should be updated with the gitignore as specified here:
https://github.com/github/gitignore/blob/master/Python.gitignore whereby venv and env and other environmental configurations are ignored. Calling the virtual enviroments  venv and env is commonly used.

The reason behind the extra_requires was to enable the developers to install at once the necessary packages to run all tests. 
We could consider to make it available as the 'contrib' option in the setup.py. Since using aiortc.contrib.signaling requires the extra packages.

setup.py
```diff
@@ -28,6 +28,14 @@
     "pylibsrtp>=0.5.6",
 ]
 
+extra_requires = {
+    'dev': [
+        'aiohttp>=3.7.0',
+        'websockets>=8.0',
```
jlaine, line 34: We need `numpy` too for `tests/test_rate.py`

jlaine:
If we're going to have a `dev` option, wouldn't it make sense to at least make use of it for our Github Actions?

jlaine:
Uhoh, somehow you've pulled the MediaBroadcaster changes into this PR!

PR #493, opened 2021-03-03
WIP: [Feature]: Multiplexing a media source

Laityned:


jlaine:
Thanks for getting the ball rolling on this, this is something I've been meaning to implement for a while.

Some questions regarding the API:

- why does the media source need to be aware of what a PeerConnection is? It needs to provide an API to create new new subscription, but adding the track to a PC is out of scope
- how do you plan to handle source with multiple tracks?
- why do we need a separate example for this? I would just update the webcam example, let's avoid repeated code.

On some more minor points could you please:
- avoid unrelated changes
- ensure the code is formatted using `black`

src/aiortc/contrib/media.py
```diff
@@ -80,7 +81,7 @@ async def stop(self):
 
 
 def player_worker(
-    loop, container, streams, audio_track, video_track, quit_event, throttle_playback
+        loop, container, streams, audio_track, video_track, quit_event, throttle_playback
```
jlaine, line 84: Unrelated change

src/aiortc/contrib/media.py
```diff
@@ -115,9 +116,9 @@ def player_worker(
 
         if isinstance(frame, AudioFrame) and audio_track:
             if (
-                frame.format.name != audio_format_name
-                or frame.layout.name != audio_layout_name
```
jlaine, line 119: Unrelated change

src/aiortc/contrib/media.py
```diff
@@ -174,9 +175,9 @@ async def recv(self):
 
         # control playback rate
         if (
-            self._player is not None
-            and self._player._throttle_playback
-            and frame_time is not None
+                self._player is not None
```
jlaine, line 178: Same here

src/aiortc/contrib/media.py
```diff
@@ -263,7 +264,7 @@ def video(self) -> MediaStreamTrack:
     def _start(self, track: PlayerStreamTrack) -> None:
         self.__started.add(track)
         if self.__thread is None:
-            self.__log_debug("Starting worker thread")
+            self.__log_debug("Starting worker __thread")
```
jlaine, line 267: .. and here

src/aiortc/contrib/media.py
```diff
@@ -284,7 +285,7 @@ def _stop(self, track: PlayerStreamTrack) -> None:
         self.__started.discard(track)
 
         if not self.__started and self.__thread is not None:
-            self.__log_debug("Stopping worker thread")
+            self.__log_debug("Stopping worker __thread")
```
jlaine, line 288: .. and here

src/aiortc/contrib/media.py
```diff
@@ -1,8 +1,10 @@
 import asyncio
+import concurrent
 import fractions
 import logging
 import threading
 import time
+from threading import Thread
```
jlaine, line 7: We already import `threading`, we don't need Thread as well

src/aiortc/contrib/media.py
```diff
@@ -382,3 +383,160 @@ async def __run_track(self, track, context):
...
+            target=self.worker,
+            args=(
+                asyncio.get_event_loop(),
+                self.__track,
+                self.__streams,
+                self.__stream_lock,
+                self.__thread_quit
+            )
+        )
+        self.__thread.start()
```
jlaine, line 412: This feels wrong, why does constructing the object start a thread?

src/aiortc/contrib/media.py
```diff
@@ -382,3 +383,160 @@ async def __run_track(self, track, context):
...
+        """
+        Stop camera thread and release camera resource
+        :return:
+        """
+        self.__thread_quit.set()
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, self.__thread.join)
+        self.__track.stop()
+
+    def add_peer_connection(self, peer_connection):
```
jlaine, line 428: This looks like it's the wrong way round. We want an API which returns the `MultiplexedMediaStreamTrack`, regardless of what we use it for.

One possible usecase: in the `server` example it would be nice to be able to take the received video and both send it to a `MediaRecorder` and send it through the "transform track" to echo it back to the browser.

src/aiortc/contrib/media.py
```diff
@@ -382,3 +383,160 @@ async def __run_track(self, track, context):
...
+        """
+        Create a new streaming track and add this to the peer connection
+        :param peer_connection:
+        :return:
+        """
+        self.__log_debug("Add track to peer connection")
+        peer_connection.addTrack(MultiplexedMediaStreamTrack(self))
+
+    @staticmethod
+    def worker(loop: asyncio.unix_events.SelectorEventLoop, track, streams, stream_lock, quit_event):
```
jlaine, line 438: Private API, prefix it with an underscore

src/aiortc/contrib/media.py
```diff
@@ -382,3 +383,160 @@ async def __run_track(self, track, context):
...
+            with stream_lock:
+                for stream in streams:
+                    futures.append(asyncio.run_coroutine_threadsafe(
+                        MultiplexedMediaSource.put_on_queues(stream, frame),
+                        loop
+                    ))
+                concurrent.futures.wait(futures)
+
+    @staticmethod
+    async def put_on_queues(stream, frame):
```
jlaine, line 466: Private API too

src/aiortc/contrib/media.py
```diff
@@ -382,3 +383,160 @@ async def __run_track(self, track, context):
...
+        :param frame: the frame that should be added to the queue
+        :return:
+        """
+        # only add frame to queue if there is any place, otherwise drop frame
+        try:
+            stream.queue.put_nowait(frame)
+        except asyncio.QueueFull:
+            logger.debug('exception on full queue')
+
+    def start_stream(self, stream):
```
jlaine, line 479: Private API

src/aiortc/contrib/media.py
```diff
@@ -382,3 +383,160 @@ async def __run_track(self, track, context):
...
+        """
+        A stream is started and calls this function so that its queue is added
+        :param stream: stream object which has a queue where frame should be dispatched to
+        :return:
+        """
+        logger.info("Add a stream to the source")
+        with self.__stream_lock:
+            self.__streams.add(stream)
+
+    def stop_stream(self, stream):
```
jlaine, line 489: Private API

Laityned:
Thank you for the constructive feedback. At this moment it is still a draft (still need to run black, add more test cases etc), so not all changes will end up in the PR.

- First point will be changed to return a new track from the source, so it is not necessary to have the peer connection
- At this moment is was created to take a track as input and not a Mediaplayer so it will be usuable with any kind of MediaStreamTrack. Using the Mediaplayer would result in using two multiplexers (one for audio, one for video).
- It will merge it with the webcam example when finished.
- This is indeed a not updated docstring yet. I had to rewrite my code to be generic (compared to the code, where you saw the logs from)


At this moment, I did not consider the use case where the received media track needed to be multiplexed. I will look into this.

jlaine:
Don't worry about the multi-track scenario for now, let's get single track right and we'll see where we go from there :)

Thanks again for tackling this, this will be a really useful addition to aiortc!

jlaine:
A point we'll need to address: I don't think "multiplexing" is the right term, as this means taking several signals and combining them into one. "Broadcasting" seems more accurate, as we are taking a single signal and sending it to multiple receivers. However the term already has a strong meaning in the field we're in, so I'm not sure what a better name would be.

@richlegrand any suggestions?

jlaine:
I've put up an alternative WIP PR at #495, I'm not sure we need threading at all here.

Laityned:
Maybe "duplicator" is a better naming?

I agree and we can remove the use of a thread. I did not realize this, when removing the webcam dependency out of this code.

PR #495, opened 2021-03-04
[media] Add a MediaRelay class

jlaine:


src/aiortc/contrib/media.py
```diff
@@ -79,6 +80,89 @@ async def stop(self):
         self.__tracks = {}
 
 
+class BroadcasterStreamTrack(MediaStreamTrack):
+    def __init__(self, broadcaster, source: MediaStreamTrack) -> None:
+        super().__init__()
+        self.kind = source.kind
+        self._broadcaster = broadcaster
+        self._queue: asyncio.Queue[Optional[Frame]] = asyncio.Queue()
```
Laityned, line 88: In case of using this track for transmitting a camera feed, it would be useful only send the latest frame and not store the complete queue.

However I could image in the case when you use it at the receiving side, you want to receive all incoming frames

examples/webcam/webcam.py
```diff
@@ -9,11 +9,34 @@
...
+broadcaster = None
+webcam = None
+
+
+def create_local_tracks(play_from):
+    global broadcaster, webcam
+
+    if play_from:
+        player = MediaPlayer(play_from)
+        return player.audio, player.video
```
Laityned, line 26: This will create a mediaplayer for each individul (starting the media source from the beginning), however I could image you want to "sync" the received media for all receiving clients

jlaine: This is on purpose: when playing from a file, we want each subscriber to start at the beginning of the file => using a MediaBroadcaster does not make sense. The only type of source for which it makes sense is for "live" sources which you can only acquire once:

- a webcam or other capture device
- a track received from a remote party

src/aiortc/contrib/media.py
```diff
@@ -79,6 +80,89 @@ async def stop(self):
...
+        if track is not None:
+            self.__log_debug("Discard proxy %s", id(proxy))
+            self.__proxies[track].discard(proxy)
+
+            # if there are no subscribers left, stop reading the source
+            if not self.__proxies[track]:
+                del self.__proxies[track]
+                if track in self.__tasks:
+                    self.__log_debug("Stop reading source %s", id(track))
+                    self.__tasks[track].cancel()
```
Laityned, line 159: Correct me if I am wrong, but in case of adding this broadcaster at the receiving side:

- Queue of RemoteStreamTrack would be able to be filling up before any proxy connected to broadcaster

- Disconnecting all streams from broadcaster, stops keep emptying the queue of RemoteStreamTrack. Reconnecting any proxy track would seem like the received stream was 'paused'.

src/aiortc/contrib/media.py
jlaine: There is a bit of a discreptancy here, because we have two different behaviours:

- for the first subscriber we only start queueing frames once recv() has been called
- for subsequent subscribers, we start queuing frames as soon as subscribe() has been called. This means during the time between subscribe() and the first call to recv() (typically the time required to establish ICE / DTLS) we are building up a backlog.

src/aiortc/contrib/media.py
jlaine: Yeah, we should probably keep reading and discarding frames if we don't want to accumulate stale frames. This makes me wonder what the correct trigger is to stop reading completely. Probably an explicit `stop()` corouting, because our MediaBroadcast is a hybrid between a source and a sink.

jlaine: .. or we just let the user call `stop()` on the source track when they are done with it, this will make the `_run()` coroutine stop, and we can perform our cleanup there.

examples/webcam/webcam.py
Laityned: Adding it as a arg option to create a "semi-live" media file source?.

src/aiortc/contrib/media.py
Laityned: Building up a backlog on transceiving is indeed undesired.

Building up a backlog (after established ICE / DTLS, in other words: not the startup backlog) on receiving in BroadcasterStreamTrack could be desired behaviour.

How about registering all proxy queues to the source on first call to recv()?

jlaine: Yup, this should be fixed in d20343e5128dd7aff651ba8c4545d8104525cf5c

src/aiortc/contrib/media.py
```diff
@@ -79,6 +80,92 @@ async def stop(self):
         self.__tracks = {}
 
 
+class BroadcasterStreamTrack(MediaStreamTrack):
+    def __init__(self, broadcaster, source: MediaStreamTrack) -> None:
+        super().__init__()
+        self.kind = source.kind
+        self._broadcaster = broadcaster
+        self._queue: asyncio.Queue[Optional[Frame]] = asyncio.Queue()
```
Laityned, line 88:
```suggestion
        if isinstance(source, RemoteStreamTrack):
            self._queue: asyncio.Queue[Optional[Frame]] = asyncio.Queue()
        else:
            self._queue: asyncio.Queue[Optional[Frame]] = asyncio.Queue(maxsize=1)
```

Tranceiver need always to send the newest frame, while receiver wants an internal buffer

jlaine: I don't want to start have hack for specific stream types. Also setting maxsize to 1 won't give you the most recent frame, but the opposite. And subsequent calls to put_nowait will bomb

jlaine: We can later add some flags to `subscribe()` which would allow us to control the queue size maybe?

Laityned: See PR on your branch

Laityned:
To come back to the naming of the new class:

i think it is more suitable to use the name MediaDuplicator instead of MediaBroadcaster to prevent confusion

jlaine:
Would `MediaRelay` fit the bill too?

Laityned:
I prefer MediaRelay over MediaRepeater, however both suits the bill

jlaine:
OK so `MediaRelay` it is. We're now left with a `subscribe()` which is maybe not super descriptive.. thoughts?

PR #498, opened 2021-03-07
[MediaRelay] Add support for unbuffered to MediaRelay

Laityned:
Support for unbuffered RelayStreamTrack, to ensure no delay between the original track and relayed track

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -395,23 +395,38 @@ class MediaRecorder:
 
 
 class RelayStreamTrack(MediaStreamTrack):
-    def __init__(self, relay, source: MediaStreamTrack) -> None:
+    def __init__(self, relay, source: MediaStreamTrack, buffered) -> None:
         super().__init__()
         self.kind = source.kind
         self._relay = relay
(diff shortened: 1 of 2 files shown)


jlaine:
This looks very cool, I'll take it for a spin. Could we add an explanation of the buffered argument to the doc string?

Laityned:
Added param to docstring

johnboiles:
I tried this out with the webcam.py example but wasn't able to see a difference in video latency. Are you seeing noticeable latency introduced by the buffering in MediaRelay?

Laityned:
On a single videostream, it is not noticable. However when using a larger number of clients, you will reach a point where your asyncio event loop is slower than framerate of the camera. 

Since the camera is in a seperate thread it keeps filling the buffer, while the clients suffer from a drop in framerate due to the performance issues of the asyncio event loop. 

The buffer will keep increasing and could lead to a out of memory. Also since the buffer is filling at a faster rate than the frames are transmitted an increased latency will occur between the client and webcam server.

johnboiles:
Got it. I experimented with it a bit more and with multiple clients connected on a Pi, it would definitely start dropping frames. This means if an I-frame gets dropped, the video gets all messed up for a bit. But probably better than OOM-ing!

PR #504, opened 2021-03-10
[Peerconnection] Add clear error when no common codec is found

Laityned:


jlaine:
I tried never to make up exceptions so far, and instead used the same names as the JS API. How does the JS API react to a lack of common codecs? I have a feeling it probably doesn't fail calling `setRemoteDescription` but instead rejects the media section in its answer.

jlaine:
Thanks for this!

PR #538, opened 2021-06-11
[peer connection] make use of remote candidates with sdpMLineIndex only

youmustrust:
## Motivation

`aiortc`'s `RTCPeerConnection` drops a remote candidate with `sdpMLineIndex` only. This is problematic if 
- the remote peer is under `ICE` lite implementation, and
- it generates a candidate with `sdpMLineIndex` only.

In this case, no agent sends out `STUN` binding request to each other, which necessarily leads to connectivity check timeout. This is an effort to make `aiortc` compatible with [`Pion`](https://github.com/pion/webrtc).

### DIFF
--- a/src/aiortc/rtcrtptransceiver.py
+++ b/src/aiortc/rtcrtptransceiver.py
@@ -80,6 +80,10 @@ class RTCRtpTransceiver:
         return self.__mid
 
     @property
+    def mline_index(self) -> Optional[int]:
+        return self.__mline_index
+
+    @property
     def receiver(self) -> RTCRtpReceiver:
         """
         The :class:`RTCRtpReceiver` that handles receiving and decoding
@@ -131,9 +135,6 @@ class RTCRtpTransceiver:
 
     def _set_mid(self, mid: str) -> None:
         self.__mid = mid
-
-    def _get_mline_index(self) -> Optional[int]:
-        return self.__mline_index
 
     def _set_mline_index(self, idx: int) -> None:
         self.__mline_index = idx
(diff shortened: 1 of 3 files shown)


youmustrust:
@jlaine Can you review this PR?

jlaine:
First of all thanks for the PR and providing a unit test!

I'm not opposed to supporting this use case, however the mline-index accessor was "marked private" (as in: starts with an underscore) on purpose: it's not defined in the specs, and don't wish library users to depend on it.

youmustrust:
Hmm, did you meant the `ORTC` spec? As for [the WebRTC spec](https://www.w3.org/TR/webrtc/#dom-rtcicecandidate-sdpmlineindex), I can clearly see it there.

jlaine:
I was referring to the change to RTCRtpTransceiver, which introduced a public `RTCRtpTranceiver.mline_index` as opposed to the private `RTCRtpTransceiver._get_mline_index()`

youmustrust:
Can you kindly elaborate on the meaning of `the specs` you said above? I mean, by which spec did you intentionally make it private.

src/aiortc/rtcrtptransceiver.py
```diff
@@ -79,6 +79,10 @@ def kind(self) -> str:
     def mid(self) -> Optional[str]:
         return self.__mid
 
+    @property
+    def mline_index(self) -> Optional[int]:
```
jlaine, line 83: This is the change that is causing me a problem.

https://www.w3.org/TR/webrtc/#webidl-1228768784 => RTCRtpTransceiver does not have an `mline_index` property

youmustrust: Aha! I see. Thanks for pointing out.

PR #559, opened 2021-09-03
feat: creating media player that works without transcoding

rprata:
I've tested `janus` example. However, I noticed that always transcode process happens, even though input codec is the same of output. I checked cpu processing consumption with transcode is ~60%.

In this PR I add an option in example that transcode does not happens. With this, cpu processing reduce to ~3% and has the same visual result. Also, latency reduce ~1.5s for video in Full HD (`1920x1080`). In this implementation, `native media player` uses h264 codec.

### DIFF
--- a/src/aiortc/codecs/base.py
+++ b/src/aiortc/codecs/base.py
@@ -2,6 +2,7 @@ from abc import ABCMeta, abstractmethod
 from typing import List, Tuple
 
 from av.frame import Frame
+from av.packet import Packet
 
 from ..jitterbuffer import JitterFrame
 
@@ -18,3 +19,6 @@ class Encoder(metaclass=ABCMeta):
         self, frame: Frame, force_keyframe: bool = False
     ) -> Tuple[List[bytes], int]:
         pass  # pragma: no cover
+
+    def pack(self, packet: Packet) -> Tuple[List[bytes], int]:
+        return bytes(packet), int(packet.pts)
--- a/src/aiortc/codecs/h264.py
+++ b/src/aiortc/codecs/h264.py
@@ -7,6 +7,7 @@ from typing import Iterator, List, Optional, Sequence, Tuple, Type, TypeVar
 
 import av
 from av.frame import Frame
+from av.packet import Packet
 
 from ..jitterbuffer import JitterFrame
 from ..mediastreams import VIDEO_TIME_BASE, convert_timebase
@@ -318,6 +319,11 @@ class H264Encoder(Encoder):
         packages = self._encode_frame(frame, force_keyframe)
         timestamp = convert_timebase(frame.pts, frame.time_base, VIDEO_TIME_BASE)
         return self._packetize(packages), timestamp
+
+    def pack(self, packet: Packet) -> Tuple[List[bytes], int]:
+        assert isinstance(packet, av.Packet)
+        packages = self._split_bitstream(bytes(packet))
+        return self._packetize(packages), int(packet.pts)
 
     @property
     def target_bitrate(self) -> int:
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -4,16 +4,17 @@ import fractions
 import logging
 import threading
 import time
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, Union
 
 import av
 from av import AudioFrame, VideoFrame
 from av.frame import Frame
+from av.packet import Packet
+from av.video.stream import VideoStream
 
 from ..mediastreams import AUDIO_PTIME, MediaStreamError, MediaStreamTrack
 
 logger = logging.getLogger(__name__)
-
 
 REAL_TIME_FORMATS = [
     "alsa",
@@ -81,7 +82,7 @@ class MediaBlackhole:
         self.__tracks = {}
 
 
-def player_worker(
+def player_worker_decode(
     loop,
     container,
     streams,
@@ -109,7 +110,7 @@ def player_worker(
     while not quit_event.is_set():
         try:
             frame = next(container.decode(*streams))
-        except (av.AVError, StopIteration) as exc:
+        except Exception as exc:
             if isinstance(exc, av.FFmpegError) and exc.errno == errno.EAGAIN:
                 time.sleep(0.01)
                 continue
@@ -153,6 +154,47 @@ def player_worker(
             asyncio.run_coroutine_threadsafe(video_track._queue.put(frame), loop)
 
 
+def player_worker_demux(
+    loop,
+    container,
+    streams,
+    audio_track,
+    video_track,
+    quit_event,
+    throttle_playback,
+    loop_playback,
+):
+    frame_time = None
+    start_time = time.time()
+
+    while not quit_event.is_set():
+        try:
+            packet = next(container.demux(*streams))
+            if not packet.size:
+                raise StopIteration
+        except Exception as exc:
+            if isinstance(exc, av.FFmpegError) and exc.errno == errno.EAGAIN:
+                time.sleep(0.01)
+                continue
+            if isinstance(exc, StopIteration) and loop_playback:
+                container.seek(0)
+                continue
+            if video_track:
+                asyncio.run_coroutine_threadsafe(video_track._queue.put(None), loop)
+            break
+
+        # read up to 1 second ahead
+        if throttle_playback:
+            elapsed_time = time.time() - start_time
+            if frame_time and frame_time > elapsed_time + 1:
+                time.sleep(0.1)
+
+        if isinstance(packet.stream, VideoStream) and video_track:
+            if packet.pts is not None and packet.time_base is not None:
+                frame_time = int(packet.pts * packet.time_base)
+                asyncio.run_coroutine_threadsafe(video_track._queue.put(packet), loop)
+
+
 class PlayerStreamTrack(MediaStreamTrack):
     def __init__(self, player, kind):
         super().__init__()
@@ -161,30 +203,33 @@ class PlayerStreamTrack(MediaStreamTrack):
         self._queue = asyncio.Queue()
         self._start = None
 
-    async def recv(self):
+    async def recv(self) -> Union[Frame, Packet]:
         if self.readyState != "live":
             raise MediaStreamError
 
         self._player._start(self)
-        frame = await self._queue.get()
-        if frame is None:
+        data = await self._queue.get()
+        if data is None:
             self.stop()
             raise MediaStreamError
-        frame_time = frame.time
+        if isinstance(data, Frame):
+            data_time = data.time
+        elif isinstance(data, Packet):
+            data_time = float(data.pts * data.time_base)
 
         # control playback rate
         if (
             self._player is not None
             and self._player._throttle_playback
-            and frame_time is not None
+            and data_time is not None
         ):
             if self._start is None:
-                self._start = time.time() - frame_time
+                self._start = time.time() - data_time
             else:
-                wait = self._start + frame_time - time.time()
+                wait = self._start + data_time - time.time()
                 await asyncio.sleep(wait)
 
-        return frame
+        return data
 
     def stop(self):
         super().stop()
@@ -230,7 +275,7 @@ class MediaPlayer:
     :param loop: Whether to repeat playback indefinitely (requires a seekable file).
     """
 
-    def __init__(self, file, format=None, options={}, loop=False):
+    def __init__(self, file, format=None, options={}, loop=False, decode=True):
         self.__container = av.open(file=file, format=format, mode="r", options=options)
         self.__thread: Optional[threading.Thread] = None
         self.__thread_quit: Optional[threading.Event] = None
@@ -238,15 +283,22 @@ class MediaPlayer:
         # examine streams
         self.__started: Set[PlayerStreamTrack] = set()
         self.__streams = []
+        self.__decode = decode
         self.__audio: Optional[PlayerStreamTrack] = None
         self.__video: Optional[PlayerStreamTrack] = None
         for stream in self.__container.streams:
             if stream.type == "audio" and not self.__audio:
-                self.__audio = PlayerStreamTrack(self, kind="audio")
-                self.__streams.append(stream)
+                if self.__decode:
+                    self.__audio = PlayerStreamTrack(self, kind="audio")
+                    self.__streams.append(stream)
             elif stream.type == "video" and not self.__video:
-                self.__video = PlayerStreamTrack(self, kind="video")
-                self.__streams.append(stream)
+                if not self.__decode:
+                    if stream.codec_context.name in ["h264", "mpeg4"]:
+                        self.__video = PlayerStreamTrack(self, kind="video")
+                        self.__streams.append(stream)
+                else:
+                    self.__video = PlayerStreamTrack(self, kind="video")
+                    self.__streams.append(stream)
 
         # check whether we need to throttle playback
         container_format = set(self.__container.format.name.split(","))
@@ -279,7 +331,7 @@ class MediaPlayer:
             self.__thread_quit = threading.Event()
             self.__thread = threading.Thread(
                 name="media-player",
-                target=player_worker,
+                target=player_worker_decode if self.__decode else player_worker_demux,
                 args=(
                     asyncio.get_event_loop(),
                     self.__container,
--- a/src/aiortc/mediastreams.py
+++ b/src/aiortc/mediastreams.py
@@ -3,10 +3,11 @@ import fractions
 import time
 import uuid
 from abc import ABCMeta, abstractmethod
-from typing import Tuple
+from typing import Tuple, Union
 
 from av import AudioFrame, VideoFrame
 from av.frame import Frame
+from av.packet import Packet
 from pyee.asyncio import AsyncIOEventEmitter
 
 AUDIO_PTIME = 0.020  # 20ms audio packetization
@@ -51,9 +52,10 @@ class MediaStreamTrack(AsyncIOEventEmitter, metaclass=ABCMeta):
         return "ended" if self.__ended else "live"
 
     @abstractmethod
-    async def recv(self) -> Frame:
+    async def recv(self) -> Union[Frame, Packet]:
         """
-        Receive the next :class:`~av.audio.frame.AudioFrame` or :class:`~av.video.frame.VideoFrame`.
+        Receive the next :class:`~av.audio.frame.AudioFrame`, :class:`~av.video.frame.VideoFrame`
+        or :class:`~av.packet.Packet`
         """
 
     def stop(self) -> None:
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -7,6 +7,7 @@ import uuid
 from typing import Callable, Dict, List, Optional, Union
 
 from av import AudioFrame
+from av.frame import Frame
 
 from . import clock, rtp
 from .codecs import get_capabilities, get_encoder, is_rtx
@@ -265,20 +266,25 @@ class RTCRtpSender:
                 pass
 
     async def _next_encoded_frame(self, codec: RTCRtpCodecParameters):
-        # get frame
-        frame = await self.__track.recv()
+        # get [Frame|Packet]
+        data = await self.__track.recv()
         audio_level = None
-        if isinstance(frame, AudioFrame):
-            audio_level = rtp.compute_audio_level_dbov(frame)
 
-        # encode frame
         if self.__encoder is None:
             self.__encoder = get_encoder(codec)
-        force_keyframe = self.__force_keyframe
-        self.__force_keyframe = False
-        payloads, timestamp = await self.__loop.run_in_executor(
-            None, self.__encoder.encode, frame, force_keyframe
-        )
+
+        if isinstance(data, Frame):
+            # encode frame
+            if isinstance(data, AudioFrame):
+                audio_level = rtp.compute_audio_level_dbov(data)
+
+            force_keyframe = self.__force_keyframe
+            self.__force_keyframe = False
+            payloads, timestamp = await self.__loop.run_in_executor(
+                None, self.__encoder.encode, data, force_keyframe
+            )
+        else:
+            payloads, timestamp = self.__encoder.pack(data)
 
         return RTCEncodedFrame(payloads, timestamp, audio_level)
 
--- a/tests/codecs.py
+++ b/tests/codecs.py
@@ -2,6 +2,7 @@ import fractions
 from unittest import TestCase
 
 from av import AudioFrame, VideoFrame
+from av.packet import Packet
 
 from aiortc.codecs import depayload, get_decoder, get_encoder
 from aiortc.jitterbuffer import JitterFrame
@@ -46,6 +47,16 @@ class CodecTestCase(TestCase):
         frame.pts = pts
         frame.time_base = time_base
         return frame
+
+    def create_video_packet(self, header, pts):
+        """
+        Create a single blank video packet.
+        """
+        buffer = header + [0]
+        packet = Packet(len(buffer))
+        packet.update(bytes(buffer))
+        packet.pts = pts
+        return packet
 
     def create_video_frames(self, width, height, count, time_base=VIDEO_TIME_BASE):
         """
--- a/tests/test_contrib_media.py
+++ b/tests/test_contrib_media.py
@@ -362,134 +362,205 @@ class BufferingInputContainer:
 
         return self.__real.decode(*args, **kwargs)
 
+    def demux(self, *args, **kwargs):
+        # fail with EAGAIN once
+        if not self.__failed:
+            self.__failed = True
+            raise av.AVError(errno.EAGAIN, "EAGAIN")
+
+        return self.__real.demux(*args, **kwargs)
+
     def __getattr__(self, name):
         return getattr(self.__real, name)
 
 
 class MediaPlayerTest(MediaTestCase):
+    def assertAudio(self, frame):
+        self.assertEqual(frame.format.name, "s16")
+        self.assertEqual(frame.layout.name, "stereo")
+        self.assertEqual(frame.samples, 960)
+        self.assertEqual(frame.sample_rate, 48000)
+
+    def assertVideo(self, frame):
+        self.assertEqual(frame.width, 640)
+        self.assertEqual(frame.height, 480)
+
+    def createMediaPlayer(self, path, **kwargs):
+        return MediaPlayer(path, **kwargs)
+
+    def endTime(self, frame):
+        return frame.time + frame.samples / frame.sample_rate
+
     @asynctest
     async def test_audio_file_8kHz(self):
         path = self.create_audio_file("test.wav")
-        player = MediaPlayer(path)
+        player = self.createMediaPlayer(path)
 
-        # check tracks
-        self.assertIsNotNone(player.audio)
-        self.assertIsNone(player.video)
+        if isinstance(self, MediaPlayerNoDecodeTest):
+            self.assertIsNone(player.audio)
+            self.assertIsNone(player.video)
+        else:
+            # check tracks
+            self.assertIsNotNone(player.audio)
+            self.assertIsNone(player.video)
 
-        # read all frames
-        self.assertEqual(player.audio.readyState, "live")
-        for i in range(49):
-            frame = await player.audio.recv()
-            self.assertEqual(frame.format.name, "s16")
-            self.assertEqual(frame.layout.name, "stereo")
-            self.assertEqual(frame.samples, 960)
-            self.assertEqual(frame.sample_rate, 48000)
-        with self.assertRaises(MediaStreamError):
-            await player.audio.recv()
-        self.assertEqual(player.audio.readyState, "ended")
+            # read all frames
+            self.assertEqual(player.audio.readyState, "live")
+            while True:
+                frame = await player.audio.recv()
+                self.assertAudio(frame)
+                if self.endTime(frame) >= 0.98:
+                    break
+            with self.assertRaises(MediaStreamError):
+                await player.audio.recv()
+            self.assertEqual(player.audio.readyState, "ended")
 
-        # try reading again
-        with self.assertRaises(MediaStreamError):
-            await player.audio.recv()
+            # try reading again
+            with self.assertRaises(MediaStreamError):
+                await player.audio.recv()
 
     @asynctest
     async def test_audio_file_48kHz(self):
         path = self.create_audio_file("test.wav", sample_rate=48000)
-        player = MediaPlayer(path)
+        player = self.createMediaPlayer(path)
 
-        # check tracks
-        self.assertIsNotNone(player.audio)
-        self.assertIsNone(player.video)
+        if isinstance(self, MediaPlayerNoDecodeTest):
+            self.assertIsNone(player.audio)
+            self.assertIsNone(player.video)
+        else:
+            # check tracks
+            self.assertIsNotNone(player.audio)
+            self.assertIsNone(player.video)
 
-        # read all frames
-        self.assertEqual(player.audio.readyState, "live")
-        for i in range(50):
-            frame = await player.audio.recv()
-            self.assertEqual(frame.format.name, "s16")
-            self.assertEqual(frame.layout.name, "stereo")
-            self.assertEqual(frame.samples, 960)
-            self.assertEqual(frame.sample_rate, 48000)
-        with self.assertRaises(MediaStreamError):
-            await player.audio.recv()
-        self.assertEqual(player.audio.readyState, "ended")
+            # read all frames
+            self.assertEqual(player.audio.readyState, "live")
+            while True:
+                frame = await player.audio.recv()
+                if self.endTime(frame) >= 1.0:
+                    break
+                self.assertAudio(frame)
+            with self.assertRaises(MediaStreamError):
+                await player.audio.recv()
+            self.assertEqual(player.audio.readyState, "ended")
 
     @asynctest
     async def test_audio_file_looping(self):
         path = self.create_audio_file("test.wav", sample_rate=48000)
-        player = MediaPlayer(path, loop=True)
+        player = self.createMediaPlayer(path, loop=True)
+
+        if isinstance(self, MediaPlayerNoDecodeTest):
+            self.assertIsNone(player.audio)
+        else:
+            # read all frames, then loop and re-read all frames
+            self.assertEqual(player.audio.readyState, "live")
+            for i in range(100):
+                frame = await player.audio.recv()
+                self.assertAudio(frame)
+
+            # read one more time, forcing a second loop
+            await player.audio.recv()
+            self.assertEqual(player.audio.readyState, "live")
+
+            # stop the player
+            player.audio.stop()
+
+    @asynctest
+    async def test_video_file_looping(self):
+        path = self.create_video_file("test.mp4", duration=5)
+        player = self.createMediaPlayer(path, loop=True)
 
         # read all frames, then loop and re-read all frames
-        self.assertEqual(player.audio.readyState, "live")
+        self.assertEqual(player.video.readyState, "live")
         for i in range(100):
-            frame = await player.audio.recv()
-            self.assertEqual(frame.format.name, "s16")
-            self.assertEqual(frame.layout.name, "stereo")
-            self.assertEqual(frame.samples, 960)
-            self.assertEqual(frame.sample_rate, 48000)
+            frame = await player.video.recv()
+            self.assertVideo(frame)
 
         # read one more time, forcing a second loop
-        await player.audio.recv()
-        self.assertEqual(player.audio.readyState, "live")
+        await player.video.recv()
+        self.assertEqual(player.video.readyState, "live")
 
         # stop the player
-        player.audio.stop()
+        player.video.stop()
 
     @asynctest
     async def test_audio_and_video_file(self):
         path = self.create_audio_and_video_file(name="test.mp4", duration=5)
-        player = MediaPlayer(path)
+        player = self.createMediaPlayer(path)
 
-        # check tracks
-        self.assertIsNotNone(player.audio)
-        self.assertIsNotNone(player.video)
+        if isinstance(self, MediaPlayerNoDecodeTest):
+            # check tracks
+            self.assertIsNone(player.audio)
+            self.assertIsNotNone(player.video)
 
-        # read some frames
-        self.assertEqual(player.audio.readyState, "live")
-        self.assertEqual(player.video.readyState, "live")
-        for i in range(10):
-            await asyncio.gather(player.audio.recv(), player.video.recv())
+            # read some frames
+            self.assertEqual(player.video.readyState, "live")
+            for i in range(10):
+                await asyncio.gather(player.video.recv())
 
-        # stop audio track
-        player.audio.stop()
+            # continue reading
+            for i in range(10):
+                await player.video.recv()
 
-        # continue reading
-        for i in range(10):
+            # stop video track
+            player.video.stop()
+
+            with self.assertRaises(MediaStreamError):
+                await player.video.recv()
+        else:
+            # check tracks
+            self.assertIsNotNone(player.audio)
+            self.assertIsNotNone(player.video)
+
+            # read some frames
+            self.assertEqual(player.audio.readyState, "live")
+            self.assertEqual(player.video.readyState, "live")
+            for i in range(10):
+                await asyncio.gather(player.audio.recv(), player.video.recv())
+
+            # stop audio track
+            player.audio.stop()
+
+            # continue reading
+            for i in range(10):
+                with self.assertRaises(MediaStreamError):
+                    await player.audio.recv()
+                await player.video.recv()
+
+            # stop video track
+            player.video.stop()
+
+            # continue reading
             with self.assertRaises(MediaStreamError):
                 await player.audio.recv()
-            await player.video.recv()
-
-        # stop video track
-        player.video.stop()
-
-        # continue reading
-        with self.assertRaises(MediaStreamError):
-            await player.audio.recv()
-        with self.assertRaises(MediaStreamError):
-            await player.video.recv()
+            with self.assertRaises(MediaStreamError):
+                await player.video.recv()
 
     @asynctest
     async def test_video_file_png(self):
         path = self.create_video_file("test-%3d.png", duration=3)
-        player = MediaPlayer(path)
+        player = self.createMediaPlayer(path)
 
-        # check tracks
-        self.assertIsNone(player.audio)
-        self.assertIsNotNone(player.video)
+        if isinstance(self, MediaPlayerNoDecodeTest):
+            self.assertIsNone(player.audio)
+            self.assertIsNone(player.video)
+        else:
+            # check tracks
+            self.assertIsNone(player.audio)
+            self.assertIsNotNone(player.video)
 
-        # read all frames
-        self.assertEqual(player.video.readyState, "live")
-        for i in range(90):
-            frame = await player.video.recv()
-            self.assertEqual(frame.width, 640)
-            self.assertEqual(frame.height, 480)
-        with self.assertRaises(MediaStreamError):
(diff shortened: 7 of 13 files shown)


ashat1701:
That's great! I also wrote something like that, but I didn't think that it so easy.

johnboiles:
This branch is working for me using the webcam.py example by simply changing `player = MediaPlayer(play_from, transcode=False)` and using a pipe to send h264 video from ffmpeg.

```
ffmpeg \
    \
    -vcodec h264 \
    -pix_fmt h264 \
    -video_size 1920x1080 \
    -i /dev/video0 \
    \
    -c:v copy \
    -f mpegts \
    pipe:1 > /run/webcam/tmp.fifo
```
```
python webcam.py --play-from /run/webcam/tmp.fifo
```
Which is fantastic because I'm able to stream low latency 1080p from my Logitech c920 on a Pi!

johnboiles:
I put my changes here: https://github.com/rprata/aiortc/pull/1

johnboiles:
I'm also not able to get it to work on Chrome, only Safari. Is that a known issue? Perhaps it's a google-really-wants-us-to-use-vp9 kinda thing?

rprata:
> I'm also not able to get it to work on Chrome, only Safari. Is that a known issue? Perhaps it's a google-really-wants-us-to-use-vp9 kinda thing?

rprata:
> This branch is working for me using the webcam.py example by simply changing `player = MediaPlayer(play_from, transcode=False)` and using a pipe to send h264 video from ffmpeg.
> 
> ```
> ffmpeg \
>     \
>     -vcodec h264 \
>     -pix_fmt h264 \
>     -video_size 1920x1080 \
>     -i /dev/video0 \
>     \
>     -c:v copy \
>     -f mpegts \
>     pipe:1 > /run/webcam/tmp.fifo
> ```
> 
> ```
> python webcam.py --play-from /run/webcam/tmp.fifo
> ```
> 
> Which is fantastic because I'm able to stream low latency 1080p from my Logitech c920 on a Pi!
> 
> Edit: I also did some profiling because even without transcoding I was still pegging the python process at 100% CPU and it was unresponsive to keyboard interrupts. Most of the time gets spent on `_split_bitstream`. One quick optimization was to stop calling `len(buf)` so many times in that method since the length of the buffer never changes. But perhaps there are other optimizations.
> 
> image

Great work! So, when you made changes in `_split_bitstream`,  how much CPU reduce? And also, is it possible to running without `ffmpeg` pipe, reading directly by `/dev/videoX`?

rprata:
> I'm also not able to get it to work on Chrome, only Safari. Is that a known issue? Perhaps it's a google-really-wants-us-to-use-vp9 kinda thing?

So, I ran `janus` example in Chrome using h264. Could explain your test in web? I don't believe about H264 codec problem in Google Chrome.

johnboiles:
> So, when you made changes in _split_bitstream, how much CPU reduce?

Looking at top it went from 100% CPU and unresponsive to keyboard interrupts to ~85% and responsive to keyboard interrupts. I'll do a more exact measurement today.

> is it possible to running without ffmpeg pipe, reading directly by /dev/videoX?

Yeah and this is what I'm doing in my PR, sorry I forgot to mention. I started with ffmpeg because I simultaneously need to decode only I-frames and save as jpeg which I can do easily in ffmpeg but I can probably figure out how to work that into the python too.

> Could explain your test in web?

I just ran the webcam.py example in my branch and opened port 8080 and hit start.

rprata:
> and

@johnboiles any transcode (decode/encode) process will increase CPU consumption. Could you confirm CPU consumption of `python` process only?

johnboiles:
> Could you confirm CPU consumption of python process only (without ffmpeg decode process)?

I'm not running the `ffmpeg` process at all anymore. I'm just running this (in my branch)

```
webcam.py --no-transcode
```

This is what I see in top without the optimization in `top`.

image

I also added some more analysis to the [`_split_bitstream` PR](https://github.com/aiortc/aiortc/pull/564). Looks like a 42% improvement in performance for this method.

johnboiles:
Sidenote: I figured out my Chrome/Firefox issue -- I needed to disable advertising of VP9. Just commented out `add_video_codec("video/VP8")` in `codecs/__init__.py`

rprata:
> > Could you confirm CPU consumption of python process only (without ffmpeg decode process)?
> 
> I'm not running the `ffmpeg` process at all anymore. I'm just running this (in my branch)
> 
> ```
> webcam.py --no-transcode
> ```
> 
> This is what I see in `top` without the optimization.
> 
> image
> 
> I also added some more analysis to the [`_split_bitstream` PR](https://github.com/aiortc/aiortc/pull/564). Looks like a 42% improvement in performance for this method.

Hi, I accept @johnboiles changes. We incorporate [PR](https://github.com/aiortc/aiortc/pull/564) here.

johnboiles:
Instead of forking the logic & classes for transcoded and untranscoded streams I wonder if you could instead rework the codebase to support `MediaStreamTrack.recv` returning `Union[av.packet.Packet, av.frame.Frame]` depending on whether transcoding is enabled or not. Then I don't think you'd need to duplicate as many classes. We would then move the `h264._split_bitstream` and `h264. _packetize` operations to to `RTCRtpSender._next_encoded_frame` since they are analogous to an `encode` operation (they turn packets/frames into bitstreams).

OR perhaps we could go even further and modify `MediaStreamTrack.recv` to ONLY operate on `av.packet.Packet`. Then for streams that need to be transcoded, both the `decode` and `encode` would happen in `RTCRtpSender._next_encoded_frame`. Not sure if that means you'd be doing a `decode` per connected client or not -- I need to experiment more. If that's the case, we're already doing an `encode` per connected client so it might be worth restructuring anyways.

Any thoughts on this?

rprata:
> Any thoughts on this?

I believe that `MediaStreamTrack.recv returning Union[av.packet.Packet, av.frame.Frame]` is a good approach. I'll update my code with your suggestion. And also, I will create a new abstract function, called `packetize()` called in `RTCRtpSender._next_encoded_frame` and implemented for each codec. In case of `h264.py`, `packetize` will contain `h264._split_bitstream` and `h264. _packetize`. What do you think about this?

johnboiles:
Yeah I think that's a solid improvement!

Zooming out for a second I see three problems we could potentially solve with a restructuring:
* Users may want to send untranscoded streams (what you handled for h264 in this PR)
* Users may want to get raw packets (for example my use case to do something special with un-transcoded I-frames
* Currently for transcoded streams there is an encode per connected user since it happens when data is being sent via rtcrtpsender.py (so even if 5 clients connect all wanting h264, the h264 stream is encoded 5 times -- unless there's some caching or something I'm missing)

I wonder if the Track objects could know how to turn themselves into bitstreams. For transcoding enabled tracks it would do an encode and return the bytes. For untranscoded tracks it might do something like packetize. Then rtcrtpsender won't care since all it gets are the bytes it needs to send out.

I haven't quite worked out what to do with MediaRelay. I think you'd need to make it configurable to either send Packets or bytes depending on the use case. Or maybe it makes both optionally available.

This also gets tricky for multiple transcoded codecs. Maybe the Player track can know how to turn itself into various encoded codecs.

rprata:
> Yeah I think that's a solid improvement!
> 
> Zooming out for a second I see three problems we could potentially solve with a restructuring:
> 
> * Users may want to send un-transcoded streams (what you handled for h264 in this PR)
> * Users may want to get raw packets (for example my use case to do something special with un-transcoded I-frames)
> * Currently for transcoded streams there is an `encode` per connected user since it happens when data is being sent via `rtcrtpsender.py` (so even if 5 clients connect all wanting the same h264, the h264 stream is encoded 5 times -- unless there's some caching or something I'm missing)
> 
> I wonder if the `*Track` objects could know how to turn themselves into bitstreams of various types. For transcoding enabled tracks it could do an encode and return the bytes. For un-transcoded tracks it might do something like `_packetize` where it turns a `Packet` into `bytes`. Then `rtcrtpsender.py` won't care since all it gets are the bytes it needs to send out for a certain format.
> 
> I haven't quite worked out what to do with `MediaRelay` in this scenario. I think you'd need to make it configurable to either send `Packet`s or `bytes` depending on the use case. This also gets tricky for multiple transcoded codecs. Maybe the `PlayerTrack` can know how to turn itself into various encoded codecs as needed, then the `MediaRelay` can actually hold multiple sets of relays, one for each output format. Then all the clients that want h264 could get an already encoded h264 stream, and all the clients that want raw bitstreams can get that. This is a pretty large change though.

@johnboiles could you review my refactoring? I made your suggestion.

johnboiles (commented):
Neat! It does the same thing but now with less code :tada:

A few suggestions on my side.

src/aiortc/codecs/h264.py
```diff
@@ -318,6 +317,13 @@ def encode(
         timestamp = convert_timebase(frame.pts, frame.time_base, VIDEO_TIME_BASE)
         return self._packetize(packages), timestamp
 
+    def pack(
+        self, packet: Packet
+    ) -> Tuple[List[bytes], int]:
+        assert isinstance(packet, av.Packet)
+        packages = self._split_bitstream(packet.to_bytes())
```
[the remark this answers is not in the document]
rprata: The packet is big for RTP. In WebRTC, `RTP` packet has a limitation of 1300bytes. I does not understand why this, however with this chunk size is the best option to play stream.

src/aiortc/mediastreams.py
```diff
@@ -144,3 +145,43 @@ async def recv(self) -> Frame:
         frame.pts = pts
         frame.time_base = time_base
         return frame
+
+class VideoStreamNativeTrack(MediaStreamTrack):
```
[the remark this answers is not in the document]
rprata: I'll remove this. It's confuse and does not work properly.

src/aiortc/contrib/media.py
```diff
@@ -100,64 +103,102 @@ def player_worker(
...
-            audio_samples += frame.samples
-
-            audio_fifo.write(frame)
-            while True:
-                frame = audio_fifo.read(audio_samples_per_frame)
-                if frame:
-                    frame_time = frame.time
-                    asyncio.run_coroutine_threadsafe(
-                        audio_track._queue.put(frame), loop
+        if transcode:
```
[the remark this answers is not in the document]
rprata: `frame` object is result of `decode()` and `packet` is result of `demux`. Because of this behavior, I need to use this condition. I don't have idea to improve this snippet.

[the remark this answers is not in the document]
rprata: I prefer to keep this code, because is safer and easy to understand (each type has a different way in condition).

src/aiortc/codecs/h264.py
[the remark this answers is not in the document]
johnboiles: Makes sense! Thanks for the explanation. I also saw this answer

https://stackoverflow.com/questions/47635545/why-webrtc-chose-rtp-max-packet-size-to-1200-bytes

And it does [look like](https://www.ietf.org/rfc/rfc3984.txt) h264 NAL units could potentially be > 1300bytes which is why we need `_packetize`. Interesting!

johnboiles (approved):

src/aiortc/mediastreams.py
```diff
@@ -51,9 +52,10 @@ def readyState(self) -> str:
         return "ended" if self.__ended else "live"
 
     @abstractmethod
-    async def recv(self) -> Frame:
+    async def recv(self) -> Union[Frame, Packet]:
         """
-        Receive the next :class:`~av.audio.frame.AudioFrame` or :class:`~av.video.frame.VideoFrame`.
+        Receive the next :class:`~av.audio.frame.AudioFrame`, :class:`~av.video.frame.VideoFrame`
+        or :class:`~av.video.frame.VideoFrame`
```
johnboiles, line 58: Nit: this should be `av.video.packet.Packet` probably

rprata: Yes. should be `av.packet.Packet`.

acschristoph:
any informations when this PR is merged ? thanks

jlaine:
There are a *lot* of changes in this PR, some of which seem to be duplicates of code in #564. Also we definitely need a CI run on this, I'm not sure why it didn't trigger?

rprata:
> There are a _lot_ of changes in this PR, some of which seem to be duplicates of code in #564. Also we definitely need a CI run on this, I'm not sure why it didn't trigger?

@jlaine I applied #564  in my branch because was a @johnboiles suggestion (you can see in history). 

About CI, can you help us trigger CI process? I just modify source code, not configuration. Is there something that I can do?

rprata:
@jlaine could you check again to start CI process. I fix the trigger problem.

src/aiortc/contrib/media.py
```diff
@@ -100,62 +109,103 @@ def player_worker(
...
 
-            frame_time = frame.time
-            asyncio.run_coroutine_threadsafe(video_track._queue.put(frame), loop)
+            # read up to 1 second ahead
+            if throttle_playback:
+                elapsed_time = time.time() - start_time
+                if frame_time and frame_time > elapsed_time + 1:
+                    time.sleep(0.1)
+
+            if isinstance(*streams, AudioStream):
```
KapJI, line 189: This doesn't work if you have both audio and video streams.
```suggestion
            if isinstance(packet.stream, AudioStream):
```

src/aiortc/contrib/media.py
```diff
@@ -100,62 +109,103 @@ def player_worker(
...
+                frame_time = int(packet.pts * packet.time_base)
+                asyncio.run_coroutine_threadsafe(audio_track._queue.put(packet), loop)
+            elif isinstance(*streams, VideoStream):
```
KapJI, line 192:
```suggestion
            elif isinstance(packet.stream, VideoStream):
```

src/aiortc/contrib/media.py
rprata: Please, could explain test to reproduce and test in my environment?

KapJI: Take some h264 encoded video and pass it to webcam example with `--play-from` and  `--no-transcode`.

rprata: So, with this suggestion worked for you?

KapJI: Yep.

rprata:
Conflicts resolved.

jlaine:
I am not sure I'm comfortable with having tracks whose `recv` method returns either Frame or Packet: these two data types are not interchangeable and you cannot blindly mix the two. Would it make more sense to have two types of tracks: regular and pre-encoded?

rprata:
> I am not sure I'm comfortable with having tracks whose `recv` method returns either Frame or Packet: these two data types are not interchangeable and you cannot blindly mix the two. Would it make more sense to have two types of tracks: regular and pre-encoded?

I did not understand how can I implement this suggestion. Could you explain more commenting in the code.

rprata:
@jlaine, I fixed tests in my branch (I believe that you need update ci timeout to 20m). 

I think that I understood now your suggestion. I could create a new abstract method called `recv_native` and keep `recv` as default (decoded).  

However I believe that is not good idea. First the default always `Frame` type (you must use `transcode=False` to change it. In the code, is there only three points to check types. The library user does not need to deal with this.

rprata:
Improve tests

rprata:
Resolved conflicts and improve tests

rprata:
@jlaine any news? I had difficult to coverage 100% of tests.

rprata:
@jlaine, I improved tests for codecov. Do you believe these modification could be merged?

Could you trigger CI again? 

Sorry for the inconvenience.

rprata:
Resolved conflicts and tested in my environment (with `pyav==9.x.x`) 
@jlaine could you approve running workflows?

src/aiortc/mediastreams.py
```diff
@@ -144,3 +146,39 @@ async def recv(self) -> Frame:
         frame.pts = pts
         frame.time_base = time_base
         return frame
+
+
+class VideoNativeStreamTrack(MediaStreamTrack):
```
jlaine, line 151: If this is purely for the test suite, move it there. Also the name is not descriptive, what does "native" mean?

src/aiortc/contrib/media.py
```diff
@@ -230,14 +280,15 @@ class MediaPlayer:
     :param loop: Whether to repeat playback indefinitely (requires a seekable file).
     """
 
-    def __init__(self, file, format=None, options={}, loop=False):
+    def __init__(self, file, format=None, options={}, loop=False, transcode=True):
```
jlaine, line 283: I'm not sure "transcode" is descriptive. The media player is not transcoding anything, it's decoding frames.

src/aiortc/contrib/media.py
```diff
@@ -161,30 +207,34 @@ def __init__(self, player, kind):
...
-        if frame is None:
+        data = await self._queue.get()
+        if data is None:
             self.stop()
             raise MediaStreamError
-        frame_time = frame.time
+        if isinstance(data, Frame):
+            data_time = data.time
+        elif isinstance(data, Packet):
+            data_time = data.dts
```
jlaine, line 222: This doesn't seem right, the time is used for a comparison against `time.time()` so it must be in seconds. The `dts` is in an arbitrary unit, never seconds.

rprata: Hi @jlaine.. you're right. In this case we need to `pts` (presentation timestamp) and convert it to time in seconds:
```
data_time = data.pts * (data.time_base.numerator / data.time_base.denominator)
```

I'll made this modification and test.

src/aiortc/contrib/media.py
rprata: I used transcode because the  original behavior is:
1 - get stream and **decode** in frames
2 - Get frames and **encode** to packet in specific codec (h264, vp8... depends

In my modification, it does not `decode->encode` frame (`transcode`). So you have a suggestion for this argument? Please, tell me a better name for this.

src/aiortc/codecs/base.py
```diff
@@ -18,3 +19,7 @@ def encode(
         self, frame: Frame, force_keyframe: bool = False
     ) -> Tuple[List[bytes], int]:
         pass  # pragma: no cover
+
+    @abstractmethod
+    def pack(self, packet: Packet) -> Tuple[List[bytes], int]:
+        pass  # pragma: no cover
```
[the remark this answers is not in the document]
rprata: I tried to follow rule of parent class (the parent class has only abstract methods). I could change this.

About te use of `dts`, we will fix this code and use `pts`. Sorry again and thank you for review.

src/aiortc/mediastreams.py
rprata: Please, tell me where I can move this code and tell me a suggestion for rename. I used `native`, because this dummy track send only dummy packets for tests.

src/aiortc/contrib/media.py
rprata: I made changes. Please, check it is ok.

src/aiortc/codecs/base.py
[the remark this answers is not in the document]
rprata: I made changes. Please, check it is ok.

src/aiortc/contrib/media.py
jlaine: Again, the MediaPlayer is not *encoding* anything, so your flag is purely controlling whether you want to perform decoding or not, i.e. if you want Packet or Frame instances in the output.

src/aiortc/mediastreams.py
jlaine: Since it's only being used by `tests/test_mediastreams.py` (unless I'm mistaken) that sounds like a good place?

src/aiortc/contrib/media.py
rprata: I'll change to `decode`. Is it ok?

src/aiortc/mediastreams.py
rprata: Ok, I'll use your suggestion.

rprata:
@jlaine 
I made all the suggestions you asked for. Could trigger the CI, again?

I agree with you that PR is huge and hard to review. But I didn't see a way to make the changes in smaller steps.

These modifications are important and I believe they should be in the main codebase, as it allows us to be able to run `aiortc` on devices with less computational resources and multiple cameras on it. I already use it in my solution and other people are forking my branch to use it too.

If you can make these changes to improve the code I would appreciate it. It's just that we've been working on it since September of last year. I believe that as it is well protected, it does not affect the default behavior and does not break the code (tests were made for this). My suggestion is to pass this solution and over time we will refactor and improve this code if necessary. I have 1 more PR to do as well (related to ice servers), but I'm focusing on fixing this one as it's very important.

jlaine:
I'm still not OK with the huge "if", in the worker, AFAICT the two branches share nothing so two separate workers would be more reable.

rprata:
@jlaine I agree with you on everything you said. And also, it is in my interest to help you maintain this project. I will refactor the code according your suggestions in small PR to be simpler to you review after merge. About `decode` parameter might be advertised as "experimental" where do you recommend put it? Would the text written below ok to advertise as experimental?

```
`decode` parameter is an experimental feature that was tested streaming with codec h264.
```

About docs errors that you said above, could you help me to fix it. I don't understand where I create this mock and how can I test this.

Feel free to commit these suggested changes.

src/aiortc/contrib/media.py
```diff
@@ -4,17 +4,19 @@
 import logging
 import threading
 import time
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, Union
 
 import av
 from av import AudioFrame, VideoFrame
+from av.audio.stream import AudioStream
```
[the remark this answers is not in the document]
rprata: I made modification in local source code (`docs/conf.py`):
```
 class MockAvFrame:
     Frame = None

+class MockAvPacket:
+    Packet = None
+
+class MockAvAudioStream:
+    AudioStream = None
+
+class MockAvVideoStream:
+    VideoStream = None

 class MockH264:
     H264Decoder = None
@@ -71,6 +78,9 @@ class MockVpx:

 sys.modules.update({'av': MockAv()})
 sys.modules.update({'av.frame': MockAvFrame()})
+sys.modules.update({'av.packet': MockAvPacket()})
+sys.modules.update({'av.audio.stream': MockAvAudioStream()})
+sys.modules.update({'av.video.stream': MockAvVideoStream()})
 sys.modules.update({'av.logging': MockAvLogging()})
 sys.modules.update({'pylibsrtp._binding': MockBinding()})
 sys.modules.update({'aiortc.codecs.h264': MockH264()})
 ````
 
 In `docs` path I ran: `make html`.  Many warnings message appears:
 
 ```
 ❯ make html
Running Sphinx v4.4.0
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 2 changed, 0 removed
reading sources... [100%] helpers
WARNING: autodoc: failed to import module 'aiortc'; the following exception was raised:
No module named 'av.audio'; 'av' is not a package
WARNING: don't know which module to import for autodocumenting 'RTCPeerConnection' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCSessionDescription' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCConfiguration' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCIceCandidate' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCIceGatherer' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCIceTransport' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCIceParameters' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCIceServer' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCCertificate()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCDtlsTransport' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCDtlsParameters()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCDtlsFingerprint()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpReceiver' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpSender' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpTransceiver' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpSynchronizationSource()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpCapabilities()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpCodecCapability()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpHeaderExtensionCapability()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpParameters()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtpCodecParameters()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRtcpParameters()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCSctpTransport' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCSctpCapabilities' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCDataChannel(transport, parameters)' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCDataChannelParameters()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'MediaStreamTrack' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCStatsReport()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCInboundRtpStreamStats()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCOutboundRtpStreamStats()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRemoteInboundRtpStreamStats()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCRemoteOutboundRtpStreamStats()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: don't know which module to import for autodocumenting 'RTCTransportStats()' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
WARNING: autodoc: failed to import module 'aiortc'; the following exception was raised:
No module named 'av.audio'; 'av' is not a package
WARNING: autodoc: failed to import class 'contrib.media.MediaPlayer' from module 'aiortc'; the following exception was raised:
No module named 'av.audio'; 'av' is not a package
WARNING: autodoc: failed to import class 'contrib.media.MediaRecorder' from module 'aiortc'; the following exception was raised:
No module named 'av.audio'; 'av' is not a package
WARNING: autodoc: failed to import class 'contrib.media.MediaBlackhole' from module 'aiortc'; the following exception was raised:
No module named 'av.audio'; 'av' is not a package
WARNING: autodoc: failed to import class 'contrib.media.MediaRelay' from module 'aiortc'; the following exception was raised:
No module named 'av.audio'; 'av' is not a package
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index
generating indices... genindex done
writing additional pages... search done
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 39 warnings.

The HTML pages are in _build/html.
 ``` 
 
 Is it right?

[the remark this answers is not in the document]
Comment (jlaine) [reply src/aiortc/contrib/media.py:11]: > I made modification in local source code (`docs/conf.py`):

Actually you don't need to worry about this anymore. I've landed a committed which completely removes the mocking of `PyAV`. This was necessary when there was no binary wheel available, but it's not relevant anymore :)

**Comment by rprata** [comment]
@jlaine news? Could you ask my latest questions?

**Comment by jlaine** [comment]
@rprata Could you please *cautiously* rebase on top of `main` and squash your commits? I'll then have a look at improving doc string.

**Comment by rprata** [comment]
@jlaine have done!

**Comment by jlaine** [comment]
I have been trying to refactor the worker + the test code, and I'm stumbling across a number of issues:

- unless I'm mistaken, none of the audio packets seem to have pts or dts
- the MediaPlayer doesn't support looping when decode=False

**Comment by jlaine** [comment]
> I don't understand about audio `pts`/`dts` issue. Is it on specific test?

Currently you don't have a test for reading a simple .wav file. If you did you would see it systematically blows up because of this code:

```
if packet.pts and packet.time_base:
    frame_time = int(packet.pts * packet.time_base)
asyncio.run_coroutine_threadsafe(audio_track._queue.put(packet), loop)
```

This check whether packet.pts and packet.time_base are "truthy" and skips calculating `frame_time` otherwise.. but still happily sends out this packet to the queue. Then the following blows up because you cannot multiply `None` by an integer:

``` data_time = data.pts * (data.time_base.numerator / data.time_base.denominator)```

.. however if you also skip sending the packet to the queue if the `pts` is None, you will hang the tests, as recv() is never going to receive anything!

> About support looping when decode=False, I want to implement this in next (PR) improvement (in refactor). So, if you want add this in PR, feel free, but it will increase complexity for this modification.

This is not a question of "increasing complexity". The loop option is there, so users will expect it to work. There is no reason for "loop" and "decode" to be incompatible, it's just that you took a different approach to exception handling compared to the "regular" worker and bail out of the worker as soon as `demux()` hits any kind of exception.

Another broken aspect is that hitting an `EAGAIN` will abort the player instead of retrying like the "normal" loop. You can see this because in your "exception_no_decode" tests you are not receiving any packets at all.

**Comment by rprata** [comment]
Hi @jlaine, 

Now I understand now. Could you help me to fix `loop and decode` behavior? 

I only tested/used my modification using H264 codec. Use a comment in docs explain these limitation is it possible for now? If not, help me to design the solution for it (to refactor this code).

**Comment by jlaine** [comment]
I'm happy to help with the refactoring, I've started by doing what I've been asking for : using 2 separate workers which I've named player_worker_decode (the existing one) and player_worker_demux (your new one). What would help is if you investigate the behaviour of PyAV / FFmpeg when reading audio files (try a mix of various formats including Opus + WAV with PCMU/PCMA codecs) : do we ever get a `pts` for such packets? If we don't we have a problem and need to either:

a/ explicitly not support audio tracks
b/ guess and populate the `pts` based on the number of audio samples we have seen so far

**Comment by rprata** [comment]
Thank you for all! I'll test PyAV behavior with `wav` audio file using codec and update here.

**Comment by jlaine** [comment]
I'm not quite done, I still need to fix the two tests which I added and are currently broken. My initial assumption that "audio doesn't give us a pts" seems to have been wrong. The problem was that we were incorrectly detecting the "end of file" condition: I thought that `demux()` would raise a `StopIteration` (like `decode()`) but apparently not, we just forever get zero-length packets.

File: src/aiortc/contrib/media.py [hunk 12]
```diff
@@ -161,30 +210,36 @@ def __init__(self, player, kind):
...
-        if frame is None:
+        data = await self._queue.get()
+        if data is None:
             self.stop()
             raise MediaStreamError
-        frame_time = frame.time
+        if isinstance(data, Frame):
+            data_time = data.time
+        elif isinstance(data, Packet):
+            data_time = data.pts * (
```
jlaine, line 225: I wonder if we couldn't just do `float(data.pts * data.time_base)`

rprata:
> I'm not quite done, I still need to fix the two tests which I added and are currently broken. My initial assumption that "audio doesn't give us a pts" seems to have been wrong. The problem was that we were incorrectly detecting the "end of file" condition: I thought that `demux()` would raise a `StopIteration` (like `decode()`) but apparently not, we just forever get zero-length packets.

I tested with 3 `.wav` files (pcm_s16le, opus and aac) using this python script:

```
import av

file = "speech_orig.wav"
in_container = av.open(
    file,
    "r",
    timeout=(5.0, 3.0),
)

audio = in_container.streams.audio

for packet in in_container.demux(audio):
    print(f"packet[pts] = {packet.pts}")
    print(f"packet[dts] = {packet.dts}")
    print(f"packet[size] = {packet.size}")
    print(f"packet[time_base] = {packet.time_base}")
    if packet.pts is None:
        print(packet)
        print("exit")
```

Only latest packet has a None `pts`. We could use the same logic used in decode =True mode.

src/aiortc/contrib/media.py
rprata: according `pyav`doc: `time_base` is a [fractions.Fraction](https://docs.python.org/3/library/fractions.html#fractions.Fraction).

I tested and it is possbile :D

src/aiortc/contrib/media.py
```diff
@@ -161,30 +210,34 @@ def __init__(self, player, kind):
...
+        elif isinstance(data, Packet):
+            data_time = float(data.pts * data.time_base)
 
         # control playback rate
         if (
             self._player is not None
             and self._player._throttle_playback
-            and frame_time is not None
+            and data_time is not None
+            and isinstance(data, Frame)
```
jlaine, line 232: Why do we check it's a frame here?

examples/webcam/webcam.py
```diff
@@ -108,6 +118,15 @@ async def on_shutdown(app):
         "--port", type=int, default=8080, help="Port for HTTP server (default: 8080)"
     )
     parser.add_argument("--verbose", "-v", action="count")
+    parser.add_argument(
+        "--preferred-codec", help="Preferred codec to use (e.g. video/H264)"
```
jlaine, line 122: This should probably be --video-codec?

examples/janus/README.rst
```diff
@@ -32,3 +32,9 @@ If you want to play a media file instead of sending green video frames, run:
 .. code-block:: console
 
    $ python janus.py --play-from video.mp4 http://localhost:8088/janus
+
+If you want to play a media without decode frames (just copying video container), run:
```
jlaine, line 36: decoding not decode

examples/janus/README.rst
```diff
@@ -32,3 +32,9 @@ If you want to play a media file instead of sending green video frames, run:
...
+
+.. code-block:: console
+
+   $ python janus.py --no-decode --play-from rtsp://user@password@ip:port/stream http://localhost:8088/janus
```
jlaine, line 40: I'm not sure I'm comfortable with this example, can `--no-decode` even work without forcing the video codec?

rprata: I believe so.

examples/webcam/webcam.py
rprata: I believe that `--preferred-video-codec` is better. Do you agree?

src/aiortc/contrib/media.py
rprata: It is not necessary more.

examples/webcam/webcam.py
jlaine: Quoting from https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences :

"Any codecs not included in codecs will not be considered during the process of negotiating a connection. This lets you prevent the use of codecs you don't wish to use."

.. so no, we are not setting a preference here, we are saying "it's this codec or nothing". This is a good thing too, because the codec is not negotiable if you're playing pre-encoded data. In fact this should probably not even be an option, if you force pre-encoded media we should detect the right codec (we can keep things simple and only support H.264, and bomb if the video track uses any other codec). I don't think we should support audio at all, because it's pretty much guaranteed to fail since our "generic" `pack` method does nothing to ensure we are sending packets which fit in UDP datagrams.

examples/janus/README.rst
jlaine: After some though I think we should focus all our efforts on making one example absolutely right, and I think that example should be the "webcam" one as it is simpler to understand and to use (you don't need a janus instance). If you're OK with this I would suggest rolling back all changes to the `janus` example and defer this for a future PR.

rprata: Ok. I tested Janus today (with mozilla and chrome browsers) with this branch and it works. So, could you test webcam example to us? Others reviews tested at the past and it was working.

Feel free to help create a new PR with changes in janus example and I help you to test (I have an environment configured here).

jlaine: The only way it can "just work" is if Janus doesn't support vpx. Depending on this doesn't seem a solid assumption.

In any case my point still stands: there is some fine tuning to be done to the examples so users don't shoot themselves in the foot (and end up opening tickets I will have to deal with). I'd rather do it in just one place until we've got it right.

rprata: Sorry @jlaine. I checked Janus Server and my video rooms  are configured to use only H264. Because of that I need to adjust this (I'll create a new PR with this).

About tickets related with this, feel free to mark me in issues to resolve related problem (I said that I help you before).

examples/webcam/webcam.py
rprata: I'll make this adjust. However, I will create new PRs to adjust this feature with each audio/video codec supported by aiortc.

rprata: > if this option is set, assert that the MediaPlayer only returns a video track containing H.264, and NO audio track
Why that is necessary to return NO audio? In this case I can remove this snippet:
```
        if isinstance(frame, AudioFrame) and audio_track:
            for frame in audio_resampler.resample(frame):
                # fix timestamps
                frame.pts = audio_samples
                frame.time_base = audio_time_base
                audio_samples += frame.samples

                frame_time = frame.time
                asyncio.run_coroutine_threadsafe(audio_track._queue.put(frame), loop)
```

rprata:
@jlaine I applied all your suggestions (docs to explain it is experimental feature, returns only H264 track and update name of option). Is it ok to be merged?

jlaine:
I'm still not enthusiastic about the changes to the examples, so in the interest of making progress here is what I suggest:

- set aside all your changes to the examples, and put them in a separate PR
- squash all the remaining changes to the actual aiortc code

That way we can get the actual code changes merged and I'll take a closer look at the examples.

rprata:
> set aside all your changes to the examples, and put them in a separate PR

Sorry, but I don't understand the steps. Could you help me? Do you want changes in examples? At the point you want revert (I implement your suggestions)?

> * squash all the remaining changes to the actual aiortc code

Do you wanna squash this into one commit only? I don't understand too.

jlaine:
Here is the patch with all your changes to the examples to create a separate PR:

```patch
diff --git a/examples/janus/README.rst b/examples/janus/README.rst
index 74a70db..17cbb12 100644
--- a/examples/janus/README.rst
+++ b/examples/janus/README.rst
@@ -32,3 +32,9 @@ If you want to play a media file instead of sending green video frames, run:
 .. code-block:: console
 
    $ python janus.py --play-from video.mp4 http://localhost:8088/janus
+
+If you want to play a media without decoding frames (just copying video container), run:
+
+.. code-block:: console
+
+   $ python janus.py --no-decode --play-from rtsp://user@password@ip:port/stream http://localhost:8088/janus
diff --git a/examples/janus/janus.py b/examples/janus/janus.py
index f913743..8c880d4 100644
--- a/examples/janus/janus.py
+++ b/examples/janus/janus.py
@@ -223,6 +223,12 @@ if __name__ == "__main__":
     parser.add_argument("--play-from", help="Read the media from a file and sent it."),
     parser.add_argument("--record-to", help="Write received media to a file."),
     parser.add_argument("--verbose", "-v", action="count")
+    parser.add_argument(
+        "--play-without-decoding",
+        help="It is an experimental feature could reduce CPU to 3%. For now it only works with video/H264 codec and no audio.",
+        action="count",
+    )
+
     args = parser.parse_args()
 
     if args.verbose:
@@ -233,7 +239,8 @@ if __name__ == "__main__":
 
     # create media source
     if args.play_from:
-        player = MediaPlayer(args.play_from)
+        decode = False if args.play_without_decoding else True
+        player = MediaPlayer(args.play_from, decode=decode)
     else:
         player = None
 
diff --git a/examples/webcam/webcam.py b/examples/webcam/webcam.py
index 2093578..4616038 100644
--- a/examples/webcam/webcam.py
+++ b/examples/webcam/webcam.py
@@ -10,6 +10,7 @@ from aiohttp import web
 
 from aiortc import RTCPeerConnection, RTCSessionDescription
 from aiortc.contrib.media import MediaPlayer, MediaRelay
+from aiortc.rtcrtpsender import RTCRtpSender
 
 ROOT = os.path.dirname(__file__)
 
@@ -18,11 +19,11 @@ relay = None
 webcam = None
 
 
-def create_local_tracks(play_from):
+def create_local_tracks(play_from, decode=True):
     global relay, webcam
 
     if play_from:
-        player = MediaPlayer(play_from)
+        player = MediaPlayer(play_from, decode=decode)
         return player.audio, player.video
     else:
         options = {"framerate": "30", "video_size": "640x480"}
@@ -65,15 +66,31 @@ async def offer(request):
             await pc.close()
             pcs.discard(pc)
 
+    decode = False if args.play_without_decoding else True
+
     # open media source
-    audio, video = create_local_tracks(args.play_from)
+    audio, video = create_local_tracks(args.play_from, decode=decode)
+
+    if video:
+        pc.addTrack(video)
+        if args.preferred_video_codec:
+            if args.play_without_decoding:
+                if args.preferred_video_codec != "video/H264":
+                    args.preferred_video_codec = "video/H264"
+            # Filter for only for the preferred_video_codec
+            codecs = RTCRtpSender.getCapabilities("video").codecs
+            preferences = [
+                codec
+                for codec in codecs
+                if codec.mimeType == args.preferred_video_codec
+            ]
+            transceiver = pc.getTransceivers()[0]
+            transceiver.setCodecPreferences(preferences)
 
     await pc.setRemoteDescription(offer)
     for t in pc.getTransceivers():
         if t.kind == "audio" and audio:
             pc.addTrack(audio)
-        elif t.kind == "video" and video:
-            pc.addTrack(video)
 
     answer = await pc.createAnswer()
     await pc.setLocalDescription(answer)
@@ -108,6 +125,15 @@ if __name__ == "__main__":
         "--port", type=int, default=8080, help="Port for HTTP server (default: 8080)"
     )
     parser.add_argument("--verbose", "-v", action="count")
+    parser.add_argument(
+        "--preferred-video-codec", help="Preferred video codec to use (e.g. video/H264)"
+    )
+    parser.add_argument(
+        "--play-without-decoding",
+        help="It is an experimental feature could reduce CPU to 3%. For now it only works with video/H264 codec and no audio.",
+        action="count",
+    )
+
     args = parser.parse_args()
 
     if args.verbose:
```

jlaine:
Merged, thanks for all the hard work and patience @rprata !

Next: let's have a PR to update one example, the `webcam` to fully leverage this new code.

PR #562, opened 2021-09-08
Implemented ability to forward an encoded stream without re-encoding a…

jpiat:
This pull request adds the ability to use a pre-encoded track (tested with H264) without re-encoding. This is especially useful for small targets that implement the H264 encoding directly in the video pipeline with zero copy to the CPU. It can also be used with webcam that outputs an h264 encoded stream, IP cameras or any embedded platform that provides only gstreamer support to access the hardware video encoder. The patch to the aiortc code does not disturb original behavior and its left to the user to implement the H264 stream parsing (see picamera example).

### DIFF
--- a/examples/picamera/README.rst
+++ b/examples/picamera/README.rst
@@ -1,0 +1,28 @@
+Webcam server
+=============
+
+This example illustrates how to use a raspberry pi and picamera to send h264 encoded video (using hardware encoder) to the browser.
+
+Running
+-------
+
+First install the required packages:
+
+.. code-block:: console
+
+    $ pip install aiohttp aiortc picamera
+
+When you start the example, it will create an HTTP server which you
+can connect to from your browser:
+
+.. code-block:: console
+
+    $ python picam.py
+
+You can then browse to the following page with your browser:
+
+http://127.0.0.1:8080
+
+Once you click `Start` the server will send video from the raspberry pi camera to the
+browser.
+
--- a/examples/picamera/pitrack.py
+++ b/examples/picamera/pitrack.py
@@ -1,0 +1,182 @@
+import asyncio
+import math
+from queue import Queue
+from struct import pack
+from typing import Iterator, List, Tuple
+
+from aiortc.mediastreams import EncodedMediaStreamTrack
+
+PACKET_MAX = 1300
+
+NAL_TYPE_FU_A = 28
+NAL_TYPE_STAP_A = 24
+
+NAL_HEADER_SIZE = 1
+FU_A_HEADER_SIZE = 2
+LENGTH_FIELD_SIZE = 2
+STAP_A_HEADER_SIZE = NAL_HEADER_SIZE + LENGTH_FIELD_SIZE
+
+
+class PiH264StreamTrack(EncodedMediaStreamTrack):
+
+    kind = "video"
+
+    _start: float
+    _timestamp: int
+
+    def __init__(self, video_rate, clock_rate=90000) -> None:
+        super().__init__()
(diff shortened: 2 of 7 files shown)


examples/picamera/pitrack.py
```diff
@@ -0,0 +1,182 @@
...
+                packetized_packages.extend(cls._packetize_fu_a(package))
+                package = next(packages_iterator, None)
+            else:
+                packetized, package = cls._packetize_stap_a(package, packages_iterator)
+                packetized_packages.append(packetized)
+
+        return packetized_packages
+
+    @staticmethod
+    def _split_bitstream(buf: bytes) -> Iterator[bytes]:
```
johnboiles, line 138: Is this the same as the implementation in h264.py? Any way to unify that code?

examples/picamera/README.rst
```diff
@@ -0,0 +1,28 @@
+Webcam server
+=============
+
+This example illustrates how to use a raspberry pi and picamera to send h264 encoded video (using hardware encoder) to the browser.
```
johnboiles, line 4: Neat! A Pi camera example seems like a great value add.

johnboiles:
Seems very nice to have a Pi camera example here!

You might want to take a look at #559 since it's a more generic way of handling video tracks that don't need to be transcoded. I'm using that change to stream a Logitech c920's native h264 stream from a Pi without transcoding. If that gets merged you might be able to simplify this change.

examples/picamera/pitrack.py
jpiat: Its the same but different  but keeping outside of the library allow the user to define it own way to split the bitstream for its own format.

jpiat:
I tried to keep the changes to aiortc as small as possible because my previous PR for this feature (almost a year ago) got rejected partly because it was to bulky to be easily reviewed. This PR only adds a new type of media track and  implement the way to process this track in rtcrtpsender. All the logic to generate the encoded track is kept outside of aiortc and is left to the user to  implement. 

I have another example ready (waiting for this PR to be accepted) which uses gstreamer to generate the h264 stream. This allow to use the C920 encoded stream but also to use IP cameras with RTSP/RTP output.

examples/picamera/pitrack.py
```diff
@@ -0,0 +1,182 @@
...
+                i += 1  # skip leading zero
+                if i + 4 >= len(buf):
+                    return
+            if buf[i] != 0 or buf[i + 1] != 0 or buf[i + 2] != 0x01:
+                i += 1
+            i += 3
+            nal_start = i
+            if (buf[nal_start] & 0x1F) != 7 and (
+                buf[nal_start] & 0x1F
+            ) != 8:  # Assuming only SPS/PPS are packaged with other NALs
```
johnboiles, line 155: So this just bundles in all the remaining NALs if the NAL at index `i` is not a SPS or PPS NAL? Why is this safe to do? Is there any reading you can point me to?

This has some performance benefits as well since you skip the rest of the method, so does it make sense to bring this into the `h264.py` version?

jpiat: This might be specific to the pi camera since the picamera encoder only output a single NAL for NALs encoding images or three NALs SPS/PPS/Image per call. Other encoders may output multiple NALs at once (a full GOP or NALs for example) so there is a need to  continue splitting the bitstream after an image NAL is discovered.

jlaine:
Is this superseeded by PR #559?

jlaine:
I'm closing this since #559 was merged, which allows a track to return `Packet` instances.

PR #564, opened 2021-09-09
Performance optimization for `_split_bitstream`

johnboiles:
In profiling performance of #559, I noticed that a ton of time is spent in `_split_bitstream`. So I set out to speed it up. In my test case (passing through un-transcoded h264 1080p), it's now 280x faster than it was. I think it's also a lot easier to read.

Here's what I'm using to profile

```
python -m cProfile -o out.profile webcam.py
snakeviz -s -H 0.0.0.0 out.profile
```

### DIFF
--- a/src/aiortc/codecs/h264.py
+++ b/src/aiortc/codecs/h264.py
@@ -222,32 +222,30 @@ class H264Encoder(Encoder):
 
     @staticmethod
     def _split_bitstream(buf: bytes) -> Iterator[bytes]:
-        # TODO: write in a more pytonic way,
-        # translate from: https://github.com/aizvorski/h264bitstream/blob/master/h264_nal.c#L134
+        # Translated from: https://github.com/aizvorski/h264bitstream/blob/master/h264_nal.c#L134
         i = 0
         while True:
-            while (buf[i] != 0 or buf[i + 1] != 0 or buf[i + 2] != 0x01) and (
-                buf[i] != 0 or buf[i + 1] != 0 or buf[i + 2] != 0 or buf[i + 3] != 0x01
-            ):
-                i += 1  # skip leading zero
-                if i + 4 >= len(buf):
-                    return
-            if buf[i] != 0 or buf[i + 1] != 0 or buf[i + 2] != 0x01:
-                i += 1
+            # Find the start of the NAL unit
+            # NAL Units start with a 3-byte or 4 byte start code of 0x000001 or 0x00000001
+            # while buf[i:i+3] != b'\x00\x00\x01':
+            i = buf.find(b"\x00\x00\x01", i)
+            if i == -1:
+                return
+
+            # Jump past the start code
             i += 3
             nal_start = i
-            while (buf[i] != 0 or buf[i + 1] != 0 or buf[i + 2] != 0) and (
-                buf[i] != 0 or buf[i + 1] != 0 or buf[i + 2] != 0x01
-            ):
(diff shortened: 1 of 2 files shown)


johnboiles:
If I'm reading this right this is a 42% improvement (using cumulative percall metric) to `_split_bitstream`. Here's without the optimization:

image

Here's with the optimiazation:

image

Cumulative time per call shows an improvement of 42%! I think this is the most relevant metric since it includes calls to sub functions.

`(0.007372-0.004275)/0.007372 = 0.42010309`

But total time is also improved by 30%

`(0.006159-0.004275)/0.006159 = 0.30589381`

johnboiles:
Not satisfied with a 40% improvement I dove deeper and got to a 2+ **order of magnitude** speed increase 🚀 😄 Apparently `bytes.find` is very efficient.

After the first commit (roughtly the same as the second pic in the last comment): 0.004232s cumulative per call

image

Now with the second commit: 0.00002629s cumulative per call (and wayyyy down in the list of functions in the profile)

image

Based on the original measurement of 0.007372s cumulative per call, that's a 99.64338% improvement or 280x faster!

We can squash these commits together when we merge, but I think they're interesting on their own to see the improvement process.

X-Ryl669:
You're modifying the behavior of the code here. In H264, NAL unit can start by either `0x00 00 01` or `0x00 00 00 01`. In your patch, you've dropped the second case. If I were you, I would check what is the byte before the `0x00 00 01` (once found) and if it's 0x00, then adjust your len and pointer accordingly. 

Nice improvement by the way.

johnboiles:
Haha yeah reading the old code it was _really_ hard to understand the cases it was handling. The resulting code is simple but it's the result of a lot of time getting confused then clarifying :) Thanks for thinking about it!

I also went ahead and fixed up the test to actually check if the right packet is output (not just the right number of packets). Looks like it was using literals wrong anyways (`b'\ff' != b'\xff'`)

jlaine:
This is awesome, thanks for the very detailed description, more readable code and improved tests!

jlaine:
I fixed the linter error, but the test suite fails. Could you check whether it's the code or the test that's wrong?

jlaine:
@rprata @johnboiles any chance of fixing the failing test (or the code)?

johnboiles:
Yeah I can just might be a couple days

jlaine:
@johnboiles any news on this, I'd love to merge it?

johnboiles:
Weird, I must have made that last part of the test without actually running it. My bad. Should be good to go now.

jlaine:
Hm, the linter error is your fault, the test against apprtc.io is not..

lgrahl:
Is the apprtc test against the live server? Because [Google decided to take it down](https://groups.google.com/g/discuss-webrtc/c/H7XuZfgkGH0/m/FR_w_VKdBgAJ).

jlaine:
> Is the apprtc test against the live server? Because [Google decided to take it down](https://groups.google.com/g/discuss-webrtc/c/H7XuZfgkGH0/m/FR_w_VKdBgAJ).

Yeah it was. I've put together a PR which rips out anything related to AppRTC in #623

jlaine:
OK well we're almost but not quite there: there is still a branch of code which has no corresponding unit test. Could you fix this please?

jlaine:
I have added the missing unit test, and am merging the PR, thanks so much!

PR #570, opened 2021-09-20
[examples] add python to python webcam example using socket.io

djsamseng:
Hey thanks for giving us this library! Just thought this additional example might be useful for those trying to get a barebones implementation together that shows most of the parts from the ground up.


examples/socketio-webcam-videostream/pycli/cli.py
```diff
@@ -0,0 +1,108 @@
...
+    else:
+        webcam = MediaPlayer("/dev/video0", format="v4l2", options=options)
+    if webcam and webcam.audio:
+        pc.addTrack(webcam.audio)
+
+    if webcam and webcam.video:
+        pc.addTrack(webcam.video)
+
+async def createOffer():
+    print("Creating an offer. Waiting for an answer")
```
rprata, line 40: Use `logging` class to print message and add `--verbose` `arg` option in your example.

jlaine:
Hi! I'm not sure I understand what this example does. Is it just changing the signaling method?

jlaine:
@djsamseng Do you think we could have a revised version of this PR which *only* provides a new signaling method, so that it can be plugged into all existing examples? Also, in 2022 what's the advantage of using socket.io as opposed to a plain WebSocket?

djsamseng:
Fair point - I guess this isn't needed then!

jlaine:
I'd still welcome an additional signal method that uses a WebSocket :)

PR #583, opened 2021-11-06
Fix webcam.py example run instruction

nowke:
Fix run instruction for playing from a media file: `server.py` to `webcam.py`


jlaine:
Thanks!

PR #592, opened 2021-11-24
connectionstatechange should be emitted when remote is closing the co…

guillaumeldc:
I couldn't figure out why I couldn't catch the "closed" state when the remote peer was closing the connection. The connectionstatechange event is not emitted for this use case. 
When the remote peer close the connection dtlsStates contains ({'closed'},) so we should also emit the connectionstatechange event.


jlaine:
Can you please provide the link to the relevant spec, I'm not sure this is correct? Also I'm surprised no tests need updating?

guillaumeldc:
https://w3c.github.io/webrtc-pc/#update-the-connection-state
Whenever the state of an RTCDtlsTransport changes **or** when the [[IsClosed]] slot turns true, the user agent MUST update the connection state. 

My understanding is that the current code only emits the closed event when the internal slot turn true.
Yes the tests might need updating, I'll take look at that.

PR #615, opened 2022-01-12
Add support for looping to MediaPlayer

juberti:
e.g., `player = MediaPlayer("foo.mp3", loop=True)` results in continuous playback. This is useful for bots and other sorts of testing.

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -82,7 +82,14 @@ class MediaBlackhole:
 
 
 def player_worker(
-    loop, container, streams, audio_track, video_track, quit_event, throttle_playback
+    loop,
+    container,
+    streams,
+    audio_track,
+    video_track,
+    quit_event,
+    throttle_playback,
+    loop_playback,
 ):
     audio_fifo = av.AudioFifo()
     audio_format_name = "s16"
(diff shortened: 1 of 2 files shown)


juberti:
@jlaine

jlaine:
This looks like a great addition, but there is a caveat: it's going to blow up for sources which are not seekable. For instance I don't expect this to work on an RTMP source.

At the very least we need this to be documented in the docstring, and (if possible) asserted in the `MediaPlayer` constructor.

juberti:
OK, PTAL. I'm having some trouble running the tests locally at the moment but will double-check this later today.

jlaine:
The linter was unhappy so I tweaked your branch, hope that's OK?

For future reference you can easily reformat the code to comply with the linter:

```pip install black
black src
```

juberti:
ah nice, thanks for doing that. Suggest adding that to the README for future contributors.

jlaine:
CI seems to hang with this PR, and I can reproduce it locally by running only `MediaPlayerTest`:

`python -m unittest tests.test_contrib_media.MediaPlayerTest`

.. and even more specifically:

`python -m unittest tests.test_contrib_media.MediaPlayerTest.test_audio_file_looping`

The behaviour is that the tests pass, but there seems to be a thread which is not shut down and the process does not exit.

jlaine:
Ah I think I have it, the new test seems to be missing:

`player.audio.stop()`

Let's see what CI has to say.

jlaine:
Merged, thanks @juberti !

PR #619, opened 2022-01-17
Calculate audio energy and send in RTP header extension

### LINKED ISSUE #614: Add support for ssrc-audio-level header extension
Currently there is some support for this in aiortc, but it's not fully wired up. I believe the required work is:
1) add "urn:ietf:params:rtp-hdrext:ssrc-audio-level" to HEADER_EXTENSIONS in __init__.py
2) return audio frame energy from `_next_encoded_frame` and pass it to `packet.extensions.audio_level`

I can take a stab at this if there's no planned work here.

juberti:
Fixes #614

### DIFF
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -259,7 +259,7 @@ class RTCRtpSender:
         frame = await self.__track.recv()
         audio_level = None
         if isinstance(frame, AudioFrame):
-            level = self._compute_audio_level_dbov(frame)
+            audio_level = self._compute_audio_level_dbov(frame)
 
         # encode frame
         if self.__encoder is None:
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import math
 import random
+import struct
 import time
 import traceback
 import uuid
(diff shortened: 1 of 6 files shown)


juberti:
@jlaine

jlaine:
Hi @juberti thanks for the PR!

Some notes:

- I've let CI run and it looks like the linters are not happy
- maybe it would make sense to pull the `_compute_level_dbov` method out to a plain function (for instance in `aiortc.rtp`) so that we can have a unit test with some test vectors?
- I see `level` is set to `None` for video frames, isn't this going to blow up: `packet.extensions.audio_level = (False, -enc_frame.level)`
- how costly is this calculation?

src/aiortc/rtcrtpsender.py
```diff
@@ -36,12 +38,19 @@
...
 
 RTP_HISTORY_SIZE = 128
 RTT_ALPHA = 0.85
 
-
+class RTCEncodedFrame:
+    def __init__(self, payloads: List[bytes], timestamp: int, level: int):
+        self.payloads = payloads
+        self.timestamp = timestamp
+        self.level = level
```
jlaine, line 52: Maybe `audio_level` would be more explicit?

juberti:
OK, fixed the lint issues and the video-frame issue, and broke out the energy calc function and added test for it. 

As for costs, it's small but not nothing, maybe 3% of a core on my system. numpy might be an option if this is too much, although there might be other costs incurred roundtripping to numpy and back.

jlaine:
I ran `isort` on the code which fixes additional linter errors.

juberti:
Thanks for doing that, I had run flake8 and black but not isort.

jlaine:
For my information, does this header need to be negotiated or are we good to always send it?

juberti:
It should be negotiated, that's right. I think there is already logic to only send the header if it's been negotiated, although it will always be computed.

jlaine:
OK, I'm going to merge this PR as is, even though I'm a bit concerned by the computational overhead (some of aiortc's usecase are on low power devices).

If issues suddenly get reported related to performance, would you consider working on a follow up PR to make this extension optional? I noticed that webrtc-extensions mentions a `RTCRtpTransceiver.setOfferedRtpHeaderExtensions` method, maybe that could be used?

jlaine:
Oops, I forgot to squash the commits, so I've had to force-push to `main`..

juberti:
Will you be publishing a new rev of aiortc with these changes?

PR #640, opened 2022-02-10
[build] produce binary wheels for linux/aarch64

jlaine:


jlaine:
@odidev you might like this one :)

odidev:
@jlaine, Thanks for your effort. It will be a great help for Linux AArch64 users. May I know what is your plan for making a new release so that Linux AArch64 wheel get deployed to PyPI?

Issue #618, opened 2022-01-16
Very poor video when encoding webcam video with H264.


relent95:
When a sender encodes webcam video with H264, the decoded images on a receiver are very poor, as can be seen at the middle browser window in the following screenshot.
![aiortc_issue_618_example](https://user-images.githubusercontent.com/13058392/151338466-e5c97f66-288f-4296-9334-6f7d312af293.png)
The screen images were being fed into the v4l2 lookback device and two local server processes(webcam.py) were sending encoded video streams to the chrome browsers.
You can see an original video source on the left, a poor video on the middle and a normal video on the right.
The right one is from the server run with the fixed **aiortc.codecs.h264.H264Encoder**(It's mentioned at the end of the writing).

I tested with the official webcam.py example, patched like the following to enforce it uses H264 instead of VP8.
(add 3 lines before calling **pc.setRemoteDescription()**)
```
async def offer(request):
    ...
    from aiortc.codecs import get_capabilities
    caps = [c for c in get_capabilities('video').codecs if c.mimeType == 'video/H264']
    pc.addTransceiver('video', direction = 'recvonly').setCodecPreferences(caps)
    await pc.setRemoteDescription(offer)
```
I am using aiortc 1.2.1 and av 8.1.0 on Ubuntu 20.04.

I analyzed the problem.
In **aiortc.codecs.h264.H264Encoder._encode_frame()**, the **frame** argument was returned by an av decoder. So its picture type should be reset before sending to an av encoder.
I found a quick-and-dirty fix like the following.
(add 1 line before calling self.codec.encode())
```
    def _encode_frame(
        ...
        frame.pict_type = av.video.frame.PictureType.NONE
        for package in self.codec.encode(frame):
        ...
```

jlaine:
Could you please elaborate I'm not sure I understand?

relent95:
I will summarize the issue. The current H264Encoder implementation does not reset picture type of input frames before calling av.CodecContext.encode(), which causes a poor video on the receiving peer.
It's a poorly documented feature of the ffmpeg library. See this stackoverflow question.[https://stackoverflow.com/questions/19456745/libavcodec-libx264-do-not-produce-b-frames]

jlaine:
OK would you care to put together a PR?

jlaine:
To make sure I understand, does this issue also affect `MediaRecorder`, which also takes frames end sends them to an encoder?

relent95:
[...]

PR #641, opened 2022-02-10
[h264] reset picture type to ensure B-frames are produced (fixes: #618)

jlaine:


src/aiortc/codecs/h264.py
```diff
@@ -277,6 +277,9 @@ def _encode_frame(
             self.buffer_pts = None
             self.codec = None
 
+        # reset the picture type, otherwise no B-frames are produced
+        frame.pict_type = av.video.frame.PictureType.NONE
```
rprata, line 281: Why necessary?

jlaine: See the discussion in the original ticket.

rprata (approved):

PR #668, opened 2022-03-15
Playout delay support

luc4s:
Hi,

First, thanks for the awesome library :)

I needed playout delay header extension support for a low latency streaming application (https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/playout-delay/README.md), so I thought I could share it.

### DIFF
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -96,8 +96,8 @@ class RTCRtpSender:
         self.__transport = transport
 
         self.__playout_delay = None
-        and self.__playout_delay_first_written > 0 self.__write_playout_delay = False
-        self.__playout_delay_first_written > -1
+        self.__playout_delay_first_written = -1
+        self.__playout_delay_write = False
 
         # stats
         self.__lsr: Optional[int] = None
@@ -229,8 +229,10 @@ class RTCRtpSender:
                         self.__rtt = RTT_ALPHA * self.__rtt + (1 - RTT_ALPHA) * rtt
 
                 # Stop writing playout delay when seen by receiver
-                if self.__write_playout_delay and self.__playout_delay_first_written > 0:
-                    self.__write_playout_delay = self.__playout_delay_first_written > report.highest_sequence
+                if self.__playout_delay_write and self.__playout_delay_first_written > 0:
+                    self.__playout_delay_write = self.__playout_delay_first_written > report.highest_sequence
+                    if not self.__playout_delay_write:
+                        self.__playout_delay_first_written = -1
 
                 self.__stats.add(
                     RTCRemoteInboundRtpStreamStats(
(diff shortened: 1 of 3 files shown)


src/aiortc/rtcrtpsender.py
```diff
@@ -221,6 +231,10 @@ async def _handle_rtcp_packet(self, packet):
                     else:
                         self.__rtt = RTT_ALPHA * self.__rtt + (1 - RTT_ALPHA) * rtt
 
+                # Stop writing playout delay when seen by receiver
+                if self.__playout_delay_write:
+                    self.__playout_delay_write = report.highest_sequence < self.__playout_delay_first_written
```
jlaine, line 236: This doesn't account for uint16 wrap-around, I think you need `uint16_gt(self.__playout_delay_first_written, self.__playout_delay_first_written)`

src/aiortc/rtcrtpsender.py
```diff
@@ -330,6 +344,12 @@ async def _run_rtp(self, codec: RTCRtpCodecParameters) -> None:
                         clock.current_ntp_time() >> 14
                     ) & 0x00FFFFFF
                     packet.extensions.mid = self.__mid
+
+                    if self.__playout_delay_write:
+                        packet.extensions.playout_delay = self.__playout_delay
+                        if sequence_number < self.__playout_delay_first_written:
```
jlaine, line 350: Likewise, this comparison is not safe

src/aiortc/rtcrtpsender.py
```diff
@@ -170,6 +174,12 @@ def replaceTrack(self, track: Optional[MediaStreamTrack]) -> None:
     def setTransport(self, transport) -> None:
         self.__transport = transport
 
+    def setPlayoutDelay(self, min, max) -> None:
+        mask = 0xFFF
+        self.__playout_delay = ((min & mask) << 12) | (max & mask)
+        self.__playout_delay_first_written = float("inf")
```
jlaine, line 180: Is this member a float or an int?

luc4s: Both, so that comparison with sequence number always pass, but I was not aware it had to deal with uint16.
So I think this won't work, I'll change it.

jlaine:
This PR has no unit test?

luc4s:
True, I haven't added any, I will add some tests.

luc4s:
Ok so I added a test for the playout delay and adapted the ones for header extensions.

Also I was wondering: regarding `setPlayoutDelay`, is raising a `ValueError` when one of the delays is out of range ok for you? 
I was also thinking of clamping them, but then I did not like having them 'silently' clamped.

luc4s:
Is there a problem with the runners? Seems like the test never started:
```
2022-04-04T13:49:27.6313834Z Requested labels: macos-latest
2022-04-04T13:49:27.6313880Z Job defined at: aiortc/aiortc/.github/workflows/tests.yml@refs/pull/668/merge
2022-04-04T13:49:27.6313901Z Waiting for a runner to pick up this job...
2022-04-04T13:49:30.3209105Z Job is waiting for a hosted runner to come online.
2022-04-04T13:49:42.6601603Z Job is about to start running on the hosted runner: GitHub Actions 4 (hosted)
```

DerGenaue:
@jlaine Is there a chance for this PR to progress some time soon?

jlaine:
@fippo I'm uncomfortable adding an API which doesn't seem to have a counterpart in native WebRTC. Do you know how this extension is used in actual browsers?

PR #672, opened 2022-03-22
webcam example does not need OpenCV

alegonz:
Hi and thank you for this great library.

As stated in the PR title, I noticed `opencv-python` is not needed to run the `webcam` example, right?

(I figured this small documentation change does not warrant to open an issue and went ahead with this PR.)

### DIFF
--- a/examples/webcam/README.rst
+++ b/examples/webcam/README.rst
@@ -11,7 +11,7 @@ First install the required packages:
 
 .. code-block:: console
 
-    $ pip install aiohttp aiortc opencv-python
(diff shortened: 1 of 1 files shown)


jlaine:
Hi @alegonz you're quite right it doesn't need OpenCV! This looks like it might be some hasty copy-and-paste.

alegonz:
@jlaine I figured that was the case, it happens to everyone 😆 

I checked the other examples and this seems to be the only one that needs the doc fix.

jlaine:
Merged, thanks!

PR #683, opened 2022-04-04
[tests] run test suite on binary wheels

jlaine:


rprata:
Why this approach? Is it necessary run tests in binary wheels or is there any benefit to this?

jlaine:
> Why this approach? Is it necessary run tests in binary wheels or is there any benefit to this?

Currently we are not testing macOS or Windows *at all*. Also since we provide wheels it's up to us to check what we are shipping is not broken. Binary wheels come with their own pitfalls, as we are bundling compiled C code.

PR #685, opened 2022-04-04
fix: adjusting regex for stun url

rprata:
I am using private stun/turn servers (e.g. Twilio). I noticed that stun url is not according with regex: `stun:global.stun.twilio.com:3478?transport=udp`. This PR is a suggestion to fix this behavior. Follow [Twilio docs for more detail](https://www.twilio.com/docs/stun-turn).

### DIFF
--- a/src/aiortc/rtcicetransport.py
+++ b/src/aiortc/rtcicetransport.py
@@ -12,6 +12,9 @@ from .rtcconfiguration import RTCIceServer
 
 STUN_REGEX = re.compile(
     r"(?P<scheme>stun|stuns)\:(?P<host>[^?:]+)(\:(?P<port>[0-9]+?))?"
+    # RFC 7064 does not define a "transport" option but some providers
+    # include it, so just ignore it
+    r"(\?transport=.*)?"
 )
 TURN_REGEX = re.compile(
     r"(?P<scheme>turn|turns)\:(?P<host>[^?:]+)(\:(?P<port>[0-9]+?))?"
--- a/tests/test_rtcicetransport.py
+++ b/tests/test_rtcicetransport.py
@@ -37,6 +37,14 @@ class ConnectionKwargsTest(TestCase):
             {"stun_server": ("stun.l.google.com", 19302)},
         )
 
+    def test_stun_with_suffix(self):
+        self.assertEqual(
+            connection_kwargs(
+                [RTCIceServer("stun:global.stun.twilio.com:3478?transport=udp")]
+            ),
+            {"stun_server": ("global.stun.twilio.com", 3478)},
+        )
+
     def test_stun_multiple_servers(self):
         self.assertEqual(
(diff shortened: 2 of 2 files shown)


jlaine:
I'm not sure this is correct, RFC 7064 doesn't seem to make any mention of a "protocol" querystring parameter for STUN URIs:

https://datatracker.ietf.org/doc/html/rfc7064

rprata:
Sorry. I'll update about `stuns`. 

Really, it does not make a sense according with section:

```
For the optional DNS discovery procedure mentioned in [Section 9 of RFC 5389](https://datatracker.ietf.org/doc/html/rfc5389#section-9), the "stun" URI scheme implies UDP as the transport protocol for SRV lookup, and the "stuns" URI scheme indicates TCP as the transport protocol
```

In this case we need adjust twilio and others STUN URLs before use in aiortc, or it is ok aiortc accept an URL with protocol sufix?

jlaine:
Let's maybe align with browser behaviour: how o Firefox and Chrome handle these URLs? We *could* just drop the query string for STUN URLs.

rprata:
Testing in Firefox, Chrome and Safari, this URL works with/without protocol query string. 

I'll update this to drop query string of URL (it is better to keep behavior according RFC 7064).

src/aiortc/rtcicetransport.py
```diff
@@ -154,7 +155,6 @@ def parse_stun_turn_uri(uri: str) -> Dict[str, Any]:
     else:
         parsed["port"] = 3478
 
-    # set transport
```
jlaine, line 157: this didn't need to change :)

tests/test_rtcicetransport.py
```diff
@@ -190,19 +198,34 @@ def test_invalid_uri(self):
     def test_stun(self):
         uri = parse_stun_turn_uri("stun:stun.services.mozilla.com")
         self.assertEqual(
-            uri, {"host": "stun.services.mozilla.com", "port": 3478, "scheme": "stun"}
+            uri,
```
jlaine, line 201: gratuitous change now that we"re not testing "protocol", ditto for the following tests

src/aiortc/rtcicetransport.py
```diff
@@ -136,6 +136,7 @@ def connection_kwargs(servers: List[RTCIceServer]) -> Dict[str, Any]:
 
 def parse_stun_turn_uri(uri: str) -> Dict[str, Any]:
     if uri.startswith("stun"):
+        uri = uri.partition("?")[0].rstrip()
```
jlaine, line 139: I'm not sure we want rstrip(), there really is no reason to have whitespace in a URL

jlaine:
Looks good to me, if you can squash your commits I'll be happy to merge it

rprata:
@jlaine done!

PR #709, opened 2022-04-26
[codecs] fix generic packetizer

jlaine:
It must return a *list* of bytes, not bytes.

@rprata


src/aiortc/contrib/media.py
```diff
@@ -293,7 +293,7 @@ def __init__(self, file, format=None, options={}, loop=False, decode=True):
                     self.__streams.append(stream)
             elif stream.type == "video" and not self.__video:
                 if not self.__decode:
-                    if stream.codec_context.name in ["h264", "mpeg4"]:
+                    if stream.codec_context.name in ["h264", "vp8"]:
```
rprata, line 296: It will fail for mpeg4 files.

rprata:
Great work. So, I'll try to update this for other codecs. 

Thank you for all!

jlaine:
> Great work. So, I'll try to update this for other codecs.
> 
> Thank you for all!

I am currently watching a video I transcoded to H.264 + OPUS, with both audio and video played without decoding :)

src/aiortc/contrib/media.py
jlaine: OK I've updated the tests

PR #710, opened 2022-04-26
[media] allow using MediaPlayer with pre-encoded Opus audio

jlaine:


rprata:
Great! I'll test it at night.

src/aiortc/codecs/g711.py
```diff
@@ -69,6 +71,10 @@ def encode(
         data = self._convert(data, SAMPLE_WIDTH)
         return [data], timestamp
 
+    def pack(self, packet: Packet) -> Tuple[List[bytes], int]:
```
rprata, line 74: It works? How can test g711 codec?

rprata:
I tested in my environment and it works. I'll review this code.

> @rprata check this one out!
> 
> Convert a file of your choice containing audio + video to the appropriate format:
> 
> `ffmpeg -i some.mp4 -codec:v libx264 -profile:v baseline -level 31 -codec:a libopus -ac 2 -f mpegts video2.ts`
> 
> You can then stream it without decoding (audio + video):
> 
> `python webcam.py --play-from video.ts --play-without-decoding --audio-codec audio/opus --video-codec video/H264`

I tested in my environment and worked (only with opus codec). I've never seem a device that delivery an audio output with `opus` codec, in general is aac. For example for video, in general camera IP has h264/h265 output.

Because of that, is it possible to add support AAC in `aiortc`? I could work in this feature before.

src/aiortc/codecs/g711.py
jlaine: I haven't yet managed to get it to work, because the payload I get are too large to fit in an RTP packet. That's why the title of this ticket is about Opus

jlaine:
I have no idea about AAC, is there even an RTP packetisation format for it or browsers that support it over webrtc?

rprata:
> I have no idea about AAC, is there even an RTP packetisation format for it or browsers that support it over webrtc?

I'll check before.

PR #732, opened 2022-06-07
can filter out ipv6 addresses

zhuker:
If you construct peer connection like below, ipv6 addresses will not be gathered
```
pc = RTCPeerConnection(use_ipv6=False)
```


fippo:
see also the discussion in https://github.com/w3c/webrtc-extensions/issues/138

PR #737, opened 2022-06-17
fix: finish audio track in exception

rprata:


jlaine:
Thanks.. I'm not sure how we missed this?

PR #757, opened 2022-08-15
Create an interface for all signaling implementations

mglettig:
This will make it easier to inject a custom signalling class.


rprata:
Could you create a example with this feature?

Issue #776, opened 2022-09-28
Slow sending messages with data channels


le-chat:
When I try to send 10 messages per second in datachannel I observe that messages are sent much slower, at 5 message per second.

I did a docker example to reproduce: https://gist.github.com/le-chat/844272e8d0f91dcbb61ba98ca635cd6b

From logs:
```
08:32:51.808407 sent №100, buffered 1750, queues: _sent 1, _outbound 0, _data_channel 50; _flight_size 35
08:33:01.895772 sent №200, buffered 3600, queues: _sent 1, _outbound 1, _data_channel 100; _flight_size 35
08:33:01.982040, 100 received: №99 at 2022-09-28 08:32:51.808277 (buffered 0)
08:33:11.973638 sent №300, buffered 5436, queues: _sent 1, _outbound 0, _data_channel 151; _flight_size 36
08:33:22.056580 sent №400, buffered 7200, queues: _sent 1, _outbound 1, _data_channel 200; _flight_size 36
08:33:22.160382, 200 received: №199 at 2022-09-28 08:33:01.895633 (buffered 0)
08:33:32.146720 sent №500, buffered 9036, queues: _sent 1, _outbound 0, _data_channel 251; _flight_size 36
08:33:42.228501 sent №600, buffered 10836, queues: _sent 1, _outbound 1, _data_channel 301; _flight_size 36
08:33:42.523312, 300 received: №299 at 2022-09-28 08:33:11.973609 (buffered 0)
08:33:52.314587 sent №700, buffered 12672, queues: _sent 1, _outbound 0, _data_channel 352; _flight_size 36
```
we can see messages are stuck on `_data_channel_queue`, though flight size is low and both `_outbound_queue` and `_sent_queue` are almost empty.

It looks like the second check in https://github.com/aiortc/aiortc/blob/1.3.2/src/aiortc/rtcsctptransport.py#L1651 erroneously breaks the loop in the very beginning and is redundant.

DerGenaue:
We have the same issue:
We try to send data for every frame which fails at 30fps

jlaine:
Isn't this a duplicate of #547 ?

DerGenaue:
This issue is about the datachannel only being able to send 10 messages per second;
the other issue is about the messages being delayed in some scenarios (aka. as soon as another message is already in flight)

PR #777, opened 2022-09-28
Fix slow datachannel sending (fixes: #776)

le-chat:
It looks like the second check in https://github.com/aiortc/aiortc/blob/1.3.2/src/aiortc/rtcsctptransport.py#L1651 erroneously breaks the loop in the very beginning and is redundant.

### DIFF
--- a/src/aiortc/rtcsctptransport.py
+++ b/src/aiortc/rtcsctptransport.py
@@ -1648,7 +1648,7 @@ class RTCSctpTransport(AsyncIOEventEmitter):
         if self._association_state != self.State.ESTABLISHED:
             return
 
-        while self._data_channel_queue and not self._outbound_queue:
+        while self._data_channel_queue:
             channel, protocol, user_data = self._data_channel_queue.popleft()
 
             # register channel if necessary


jlaine:
Interesting, is there any way we can have an explicit test which captures this condition?

jlaine:
I *think* the logic here was to avoid blindly pulling data from the DataChannel down to the SCTP buffers, otherwise the datachannel layer has no idea there is a backlog further down the stack.

Furthermore, `_data_channel_flush` should be called again once the remote party acknowledges some data with a SACK.

I don't think this change should go in as-is.

le-chat:
Does not check in `RTCSctpTransport._transmit` for `_flisght_size` (https://github.com/aiortc/aiortc/blob/1.3.2/src/aiortc/rtcsctptransport.py#L1542) provide such «idea there is a backlog further down the stack»?

And is it necessary to have one SACK for each datachannel message?

jlaine:
> Does not check in `RTCSctpTransport._transmit` for `_flisght_size` (https://github.com/aiortc/aiortc/blob/1.3.2/src/aiortc/rtcsctptransport.py#L1542) provide such «idea there is a backlog further down the stack»?

The line you point to determines when we take data from the SCTP buffer and send it out on the network. A simplified picture is:

`(DATA CHANNEL buffer) => (SCTP buffer) => network`

The API user only interacts with the data channel, so their only visible metric is `RTCDataChannel.bufferedAmount`. If we unconditionally pull data from the data channel and shove it into the SCTP buffer, the API user will have no information at all: no data will be buffered at the data channel layer.

> And is it necessary to have one SACK for each datachannel message?

SACK is an SCTP concept, not datachannel. However you do have a point here: SACK is a *selective* ACK, there is no requirement that each chunk should be immediately acknowledged, only *eventually* acknowledged within 500ms of reception:

https://datatracker.ietf.org/doc/html/rfc2960#section-6.2

Anyway back to the PR: we need to dig a bit further and understand why your PR leads to an improvement. Instead of completely removing `not self._outbound_queue`, what happens if you replace it by something like `len(self._outbound_queue) < SOMENUMBER`?

Also can we agree on a specific benchmark?

jlaine:
Thanks for this great summary! As you can probably tell, the various queues try to separate the "layers" (DataChannel vs SCTP), as it's likely SCTP might not be the only transport in the future for data channels.

>     * or maybe we need call `RTCSctpTransport._data_channel_flush()` when `_outbound_queue` becomes empty, i.e. after `RTCSctpTransport._transmit()` (looks the best: checks for `_flight_size` and for `not self._outbound_queue` ensure that messages are kept in `_data_channel_queue` when network is busy, but latency is minimal when sctp layer comes back again)

OK this is interesting, this means we have no feedback loop for the SCTP layer to inform the DataChannel layer that it can try pushing more data down? This looks like a good place to attempt a change. The "workaround" I suggested earlier is really a hack, as there is probably never going to be a "right" threshold on `_outbound_queue` since it's a number of "chunks" not an actual data size.

DerGenaue:
I did some local testing and in my tests, `transmit` only ever got called from `_receive_sack_chunk()` because when in the `_send()` function, `_t3_handle` is basically always active, so transmit doesn't get called.
This means that new messages only got flushed out of the `_outbound_queue` whenever feedback arrived and not if new messages are queued and there is still space free on the wire.

Suggested fix: in `rctsctptransport.py` `_send(..` (`~L.1330`) remove the `if` and instead always call `self._transmit()`.
As far as I understand it, `_t3_handle` timer is meant to handle abortion & retransmits of non-acknowledged data chunks and NOT meant to lock down transmission of data.
`_transmit()` internally handles the maximum amount of in-flight packages and if that overflows, the `_outbound_queue` will not get cleared and thus the `_data_channel_queue` also will start filling up and all the feedback will work like intended.

This modification has the additional benefit of sending out messages immediately if there is still network available without waiting for the next feedback compared to the `Flush` patch.

`_flight_size` being several hundreds is completely ok; it is limited to `self._cwnd = 3 * USERDATA_MAX_LENGTH` which is equal to `3600` bytes.
This limits the data channel by transmitted bytes instead of number of messages, which is correct imo

le-chat:
I've added suggested by DerGenue patch to my gist (you may run `docker run -it --rm aiortc-report ./run.sh local always-transmit` or `docker run -it --rm aiortc-report ./run.sh janus always-transmit`) and from these tests it looks really like the best solution. I will try it with our main application later.

le-chat:
I tried DerGenue's patch with our application code yesterday and though it seems to improve a situation, sometimes I am still getting those `Consent to send expired` messages, after which datachannel closes. Not sure if this relates only to discussed problem, may be it is (also) a bug in `aioice` (because in logs I see consent response arrived timely, much earlier than in 30 second interval mandated by RFC 7675).

le-chat:
In our application it usually takes 12–15 minutes, my previous minimal reproducible example does not seem to capture this problem with your patch, only a full application.

Alas, I don't have more time for further investigation and am going to use another transport.

DerGenaue:
I think I was able to reproduce your issue in a slightly different form:
The sample code has a ping feature where it regularly measures the datachannel ping and that slowly increases over time, even on a local host from ~2ms to 2000ms after some time.
Probably that cumulates until it triggers the Consent expiry problem

DerGenaue:
Turns out what I saw was a fluke where the JS test code printed the ping to a ever increasing log and the print between fetching the timestamp and sending the message to the network was what created the ever-increasing ping times 🙈 

Now at least I know how to open up WebRTC connections in Wireshark (PRs incoming)...

I'm currently only able to reproduce the Consent Expiry Problem by overloading the client with messages (10k per second).
I'll investigate this a bit more to see if this could be the cause

le-chat:
It may be related to https://github.com/aiortc/aioice/issues/58 or something else in aioice. In logs I have seen that consent confirmations were arriving in time but aiortc still thought it is expired. From the code it is not obvious whether aioice accepts confirmations with older transaction id than it last sent (RFC allows this, and janus seems to do so).

DerGenaue:
How much load is there on your system?
For me it was the case that the single-threaded python gets overloaded and then the send buffers become huge.
The code that sends out consent stuff throws it on the pile, but it never makes it to the network because there are a thousand messages queued before it.
Inbound messages get through within milliseconds, though.
Be aware that the other system does also request consent; so this goes both ways.
Are you sure the "confirmations" belong to the transactions aiortc sent?
I hacked in a small log in the request method so I can see how long each took to get confirmed.
Consent Transactions currently time out after 500ms as far as I can see; so that is quite short and responses that come later get ignored.

jlaine:
I have merged #795 which I am much more comfortable with.

PR #783, opened 2022-10-16
warn users webcam.py example may not work with firefox on local network

jensbjorgensen:
comment-only


jlaine:
If you really want users to see this warning, wouldn't the example's README be a better place?

jensbjorgensen:
sure, if you prefer. I'm not sure I read the README myself ;-). branch is updated

jlaine:
Thanks for your patience, I finally got round to merging this.

Issue #547, opened 2021-07-15
datachannel has a great latency


tsonglew:
# My Code 

https://github.com/tsonglew/aiortc-datachannel-delay

run `python main.py` to start the server, then visit `localhost:8080` with a web browser

# My Problem

I'm trying to process the video frames from video track, and send back the result with datachannel. 

the result is replaced with `time.time()` in the above demo repo as following: 
``` python
self.channel.send(
      json.dumps({"now": time.time() * 1000})
)
```

the `<client receive from datachannel time> - <datachennl.send() time>` is considered as delay, and are `console.log`ed with code:
```javascript
ch.addEventListener("message", function (evt) {
	console.log(Date.now() - JSON.parse(evt.data).now);
});
```

the datachannel config is `{ordered: false, maxRetransmits: 0}`

# Output

<img width="198" alt="image" src="https://user-images.githubusercontent.com/14333766/125766778-698f9aee-2418-4fd5-ab81-c21b5be1e8fb.png">

as illustrated in the sreenshot, the delay is growing, but video and audio are fluent.

How can I get rid of the growing delay? Thx

tsonglew:
found this problem is due to `RTCSctpTransport._transmit`, solved this by calling `RTCSctpTransport._send_chunk` to send data fragments directly

tsonglew:
as a workaround, I altered my datachannel sending code like this

```python
self.channel.send(json.dumps({"now": time.time() * 1000}))
await self.channel._RTCDataChannel__transport._data_channel_flush()
await self.channel._RTCDataChannel__transport._transmit()
```

jlaine:
Could you elaborate on this please ?

tsonglew:
For apps sensitive to datachannel messages delay, the data should be transmitted immediately after calling `datachannel.send()`, but it isn't facts.

 Delay on datachannel could be reproduced in this demo: https://github.com/tsonglew/aiortc-datachannel-delay. Calling `RTCSctpTransport._data_channel_flush` and `RTCSctpTransport._transmit()` explicitly is a workaround.

I'm not familiar with aiortc and webrtc, but I guess there are two reasons:
1. user data are not tranmitted immediately after `RTCDataChannel.send`. The data are appended to `RTCSctpTransport._datachannel_queue` at first, and will be poped when `Future` `_data_channel_flush` comes, which means asyncio coroutine schduler might call this future after a period of time, and leads to the delay.
[...]

PR #795, opened 2022-11-19
Always try to transmit data messages immediately. Fixes #776 and #547

### LINKED ISSUE #776: Slow sending messages with data channels

DerGenaue:
Instead of waiting for the next reception confirmation / empty in-flight queue Fixes aiortc#776, fixes aiortc#547
Superseeds #777

In my opinion, this is the correct way to fix the datachannel latency problems.

Without this modification, sending messages appends them to the buffer, but does not transmit them if Timer 3 `_t3_handle` is active (aka. there currently are other messages in-flight).  
`_transmit` will only get called once a reception confirmation for older messages is received.

As far as I understand it, `_t3_handle` timer is meant to handle abortion & retransmits of non-acknowledged data chunks and NOT meant to lock down transmission of data.
`_transmit()` internally handles the maximum amount of in-flight packages and if that overflows, the `_outbound_queue` will not get cleared and thus the `_data_channel_queue` also will start filling up and all the feedback (eg. `RTCDataChannel.bufferedAmount`) will work like intended.

### DIFF
--- a/src/aiortc/rtcsctptransport.py
+++ b/src/aiortc/rtcsctptransport.py
@@ -1324,8 +1324,7 @@ class RTCSctpTransport(AsyncIOEventEmitter):
             self._outbound_stream_seq[stream_id] = uint16_add(stream_seq, 1)
 
         # transmit outbound data
-        if not self._t3_handle:
-            await self._transmit()
+        await self._transmit()
 
     async def _send_chunk(self, chunk: Chunk) -> None:
         """


jlaine:
I am a *lot* more comfortable with this than with #777, thanks so much!

PR #796, opened 2022-12-08
Implement force_keyframe parameter in h264

sirf:
The `force_keyframe` parameter was ignored in the `_encode_frame()` method in `h264.py`

This change sets the picture type to `I` when `force_keyframe` is `True`


jlaine:
Can you please add a unit test for this?

jlaine:
I have added the missing test, let's see if CI completes.

PR #797, opened 2022-12-09
Support `SSLKEYLOGFILE` environment variable

DerGenaue:
This allows using eg. Wireshark to debug the encrypted connection

This environment variable apparently is pretty standard in many ssl / tls scenarios such as browsers.  
It gives a path to a file, eg. `tls.keylog`, where all encryption keys are logged.

The resulting file can be plugged into the TLS module of Wireshark which then allows it to decrypt the packages it captured.

Please note that Wireshark needs additional tweaks to recognize DTLS & SCTP inside


src/aiortc/rtcdtlstransport.py
```diff
@@ -48,6 +48,10 @@
 
 assert lib.OpenSSL_version_num() >= 0x10002000, "OpenSSL 1.0.2 or better is required"
 
+# Log TLS secrets to a file, similar to browsers
+SSLKEYLOGFILE = os.getenv('SSLKEYLOGFILE')
+if SSLKEYLOGFILE:
+    logger.warning('Logging all TLS keys to %s', SSLKEYLOGFILE)
```
jlaine, line 54: Why are we emitting a warning here?

jlaine:
> Please note that Wireshark needs additional tweaks to recognize DTLS & SCTP inside

I'm not sure I understand this comment. Is the produced log file usable in Wireshark or not?

src/aiortc/rtcdtlstransport.py
DerGenaue: Because logging all encryption keys is something you probably don't want by default, so I thought it might be a good idea to log a warning if it is enabled.

DerGenaue:
Yes, the log file is perfectly usable; Wireshark just was unable to decode the protocols
I have reported the problem with SCTP over DTLS with Wireshark and they fixed it so that you can at least manually specify the protocol combination. Automatic detection doesn't work yet, though.
https://gitlab.com/wireshark/wireshark/-/issues/18719

jlaine:
Please note that #813 is going to clash with this, is it possible to enable this logging using PyOpenSSL's API?

DerGenaue:
ah, perfect, you already found it

jlaine:
PR #813 has been merged if you'd care to rework your pr on top of `main`?

DerGenaue:
I'll do it tomorrow 👍

jlaine:
> I'll do it tomorrow +1

Sure take your time, I'm not in a position to ask for immediate turnarounds ATM :)

Thanks again for your PR for the SCTP stack, the issue it fixed had been outstanding for a while.

DerGenaue:
Rebased ✅ 
Still needs testing

DerGenaue:
ok; linter is happy now

jlaine:
I'm a bit spooked by the fact coverage has not changed, who is setting the environment variable in CI?

nturay94:
Hi,

I have built a WebRTC framework based on aiortc and interested in decrypting the WebRTC data channel application payload. Is there still any ambition to include this in the main modul. If not, did you found a workaround ?

Issue #800, opened 2022-12-28
Weird behavior while trying to connect to livekit


dguerizec:
Hello,

I'm developing a python client SDK for livekit, and I ran into a weird problem.

In rctpeerconnection.py, in the __connect() method, I have to patch line 1018 to disable SCTP to make it work.

As far as I know, SCTP is used for datachannels. I don't need to use the datachannel, but livekit sends it in its initial offer.

The problem I'm seeing, is that the sctp transport is being connecting, somehow preventing the DTLS/RTP transport to connect.

I'm not sure my explanation is clear, since I'm very new to webrtc, but I've put an example here:
https://github.com/dguerizec/livekit-client-sdk-python

In this repo, the subscriber example only works when I disable SCTP in RTCPeerConnection.__connect().

If I don't disable SCTP, the dtlsTransport.state is always "connecting", thus never receiving from the transceiver.

Is there something I'm doing wrong ?

Initial offer from livekit-server:
```
v=0
o=- 4261650069995738642 1672227076 IN IP4 0.0.0.0
s=-
t=0 0
a=fingerprint:sha-256 5F:CF:73:AD:E0:DE:D0:8B:7A:B4:FF:38:57:7A:3F:BD:99:1B:82:CB:04:DB:A4:50:C3:04:1C:0B:17:BC:47:9A
a=ice-lite
a=extmap-allow-mixed
a=group:BUNDLE 0
m=application 9 UDP/DTLS/SCTP webrtc-datachannel
c=IN IP4 0.0.0.0
a=setup:actpass
a=mid:0
a=sendrecv
a=sctp-port:5000
a=ice-ufrag:qJHytwgzRTlkmPiM
a=ice-pwd:yXckeNWryDvwkHPMQXcOaZrPsASfqaGo
```

I can provide other debug information if needed, along with network traces.

jlaine:
You're going to need to provide more information to convince me `aiortc` is doing anything wrong here. All I have is an initial SDP offer and "then something doesn't work".

dguerizec:
OK, I'll try to find some time to get a minimal example showing the problem with only aiortc.
Basically, when there are multiple offers, and the first offer only contains a webrtc data channel, the subsequent offers will not take any other A/V channel into account.

PR #802, opened 2022-12-29
[issues] fix for connection to livekit (webrtc-datachannel coming first in sdp)

dguerizec:
This PR is to fix the problem described in #800

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,3 +75,7 @@ select = [
 
 [tool.setuptools.dynamic]
 version = {attr = "aiortc.__version__"}
+
+[tool.setuptools.packages.find]
+exclude = ["_cffi_src"]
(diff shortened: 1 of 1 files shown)


jlaine:
Could you please be more explicit about what you are doing here? Is livekit conforming to specs?

dguerizec:
I'm not sure which one of livekit or aiortc is compliant or not, but livekit starts by offering a single webrtc data channel, and only after the answer (from aiortc) it offers A/V streams.
I noticed that aiortc couldn't open any A/V stream after receiving the first offer containing only the data channel.
This patch fixes it, by delaying the SCTP channel establishment until at least an A/V channel is available.
I guess there may be side effects, especially if webrtc is used only for a single data channel (it will hang), but I don't know how to do better.

Issue #806, opened 2023-01-03
AttributeError: 'builtins.Certificate' object has no attribute '_x509'


abmuslim:
I am working on the WebRTC application. It is working fine on other Linux versions. But when I installed it in OpenSuse, it give me this error related to certificates.

### **Error**

INFO:aiohttp.access:185.134.139.187 [03/Jan/2023:10:14:09 +0000] "GET / HTTP/1.1" 200 2304 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36"
INFO:aiohttp.access:185.134.139.187 [03/Jan/2023:10:14:09 +0000] "GET /client.js HTTP/1.1" 200 6577 "http://193.25.X.X/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36"
ERROR:aiohttp.server:Error handling request
Traceback (most recent call last):
  File "/home/sysadmin/.local/lib/python3.10/site-packages/aiohttp/web_protocol.py", line 433, in _handle_request
    resp = await request_handler(request)
  File "/home/sysadmin/.local/lib/python3.10/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/home/sysadmin/abdullah/webrtc-server/src/server.py", line 105, in return_offer
    pc.addTransceiver(trackOrKind="video")
  File "/home/sysadmin/.local/lib/python3.10/site-packages/aiortc/rtcpeerconnection.py", line 471, in addTransceiver
    return self.__createTransceiver(
  File "/home/sysadmin/.local/lib/python3.10/site-packages/aiortc/rtcpeerconnection.py", line 1067, in __createTransceiver
    dtlsTransport = self.__createDtlsTransport()
  File "/home/sysadmin/.local/lib/python3.10/site-packages/aiortc/rtcpeerconnection.py", line 1044, in __createDtlsTransport
    dtlsTransport = RTCDtlsTransport(iceTransport, self.__certificates)
  File "/home/sysadmin/.local/lib/python3.10/site-packages/aiortc/rtcdtlstransport.py", line 375, in __init__
    self.__ctx = certificate._create_ssl_context()
[...]

PR #809, opened 2023-01-14
Fix incompatibility issue with cryptography v39

chschnell:
Added member _openssl_cert to class RTCCertificate (an OpenSSL.crypto.X509 instance derived from cryptography.x509.Certificate instance RTCCertificate._cert). Added pyOpenSSL import statement.
Replaced self._cert._x509 with self._openssl_cert._x509.

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -17,6 +17,7 @@ from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
 from pyee.asyncio import AsyncIOEventEmitter
 from pylibsrtp import Policy, Session
+from OpenSSL.crypto import X509
 
 from . import clock, rtp
 from .rtcicetransport import RTCIceTransport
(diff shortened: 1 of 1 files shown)


jlaine:
Hi! Thanks for this. It goes some way towards fixing the issue, but it still makes use of a private member (this time in pyopenssl) so it will probably eventually break. It also seems you did not add the required dependency on pyopenssl in setup.py?

chschnell:
I've added the pyopenssl dependency, I chose the current version 23.0.0 as the minimum since I don't know how far back this is compatible.

As for your second point regarding the ongoing private member use, I agree, please see my lengthy comment I posted to the issue [here](https://github.com/aiortc/aiortc/issues/806#issuecomment-1382849961). But I think there are two problems, one is using an outdated cryptography package and the other is using private members in foreign modules. To me personally fixing the first has more urgency than the latter.

So I looked for a way to fix the first problem with minimal changes to your codebase and created this PR.

I will however look into it again now to see if there is a middle way to remove using `_x509` without porting everything to pyopenssl. I would be willing to give it a try one way or another. What's your opinion on that? Would you prefer a full port to pyopenssl (to me this looks as the most robust solution for the future)?

I'd also like to apologize for my obvious lack of skills using github, I've already created a little mess with my PR here and I'm sorry about that.

avizipi:
I also have this problem, waiting for pr closing for solution.

jlaine:
I have submitted a more ambitious approach in #813 which is a proper switch to PyOpenSSL.

PR #813, opened 2023-01-22
[dtls] Use PyOpenSSL's API instead of low-level APIs

jlaine:
Until now we used a mix of `crytography` and direct calls to the OpenSSL bindings, which eventually broke. We now try to use only PyOpenSSL's Python API, with the exception of DTLS timeout handling which is not yet exposed by PyOpenSSL.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ install_requires = [
     "google-crc32c>=1.1",
     "pyee>=9.0.0",
     "pylibsrtp>=0.5.6",
+    "pyopenssl>=23.0.0",
 ]
 
 extras_require = {
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -1,5 +1,4 @@
 import asyncio
-import base64
 import binascii
 import datetime
 import enum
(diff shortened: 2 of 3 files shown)


jlaine:
For a complete switch to PyOpenSSL, we would need wrappers for DTLSv1_get_timeout and DTLSv1_handle_timeout. I have submitted the following PR: https://github.com/pyca/pyopenssl/pull/1180

chschnell:
In `certificate_digest()` I need to pass "SHA256" instead of "SHA-256" as the argument to `digest()`, otherwise I get an error "ValueError: No such digest method".

jlaine:
That's weird, how come CI passes?

chschnell:
I don't know, but note that it also used to be "SHA256" when interfacing openssl directly, before this commit it was:

    digest = lib.EVP_get_digestbyname(b"SHA256")

After changing it back to "SHA256" my test application runs perfectly fine (Raspberry Pi OS on Raspi Zero 2W).

chschnell:
Here's how to make openssl list the supported digest method names (I've cut out most of them):

    $ openssl list -digest-algorithms
    ...
    SHA224
    sha224WithRSAEncryption => SHA224
    SHA256
    sha256WithRSAEncryption => SHA256
    SHA3-224
    ...

jlaine:
It looks like you're right:

```python
>>> from OpenSSL._util import lib
>>> lib.EVP_get_digestbyname(b"SHA256")
<cdata 'EVP_MD *' 0x7fb9ce8ab100>
>>> lib.EVP_get_digestbyname(b"SHA-256")
<cdata 'EVP_MD *' NULL>
```

.. and yet when I run my unit tests it doesn't break?! Anyway I have submitted #814

PR #831, opened 2023-03-01
Show start button after clicking Stop in server example

mgraczyk:


mgraczyk:
This change makes it possible to stop and restart the demo.

jlaine:
The "start" button is purposefully disabled, as allowing a restart means fully resetting the state.

mgraczyk:
Thanks

PR #871, opened 2023-04-24
Add webcam selection list for ./examples/server/ example

Combinacijus:
image

Developed as a workaround to test Realsense camera which first webcam is depth image in Y16 format which can't be sent
```
[10076:11:0424/164151.763473:ERROR:webrtc_video_track_source.cc(160)] We cannot send frame 
with storage type: format: PIXEL_FORMAT_Y16 storage_type:UNOWNED_MEMORY 
coded_size:640x480 visible_rect:0,0 640x480 natural_size:640x480 timestamp:22190317
[10076:11:0424/164151.763808:ERROR:(-1)] Check failed: false.
```

### DIFF
--- a/examples/server/client.js
+++ b/examples/server/client.js
@@ -166,14 +166,23 @@ function start() {
         if (constraints.video) {
             document.getElementById('media').style.display = 'block';
         }
-        navigator.mediaDevices.getUserMedia(constraints).then(function(stream) {
-            stream.getTracks().forEach(function(track) {
-                pc.addTrack(track, stream);
+
+        var deviceId = document.getElementById('camera-selection').value;
+        constraints.video = {
+            deviceId: { exact: deviceId }
+        };
+
+        navigator.mediaDevices.getUserMedia(constraints)
+            .then(function(stream) {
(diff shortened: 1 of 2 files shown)


examples/server/client.js
```diff
@@ -166,14 +166,23 @@ function start() {
         if (constraints.video) {
             document.getElementById('media').style.display = 'block';
         }
-        navigator.mediaDevices.getUserMedia(constraints).then(function(stream) {
-            stream.getTracks().forEach(function(track) {
-                pc.addTrack(track, stream);
+
+        var deviceId = document.getElementById('camera-selection').value;
+        constraints.video = {
```
jlaine, line 171: You are nuking all the settings which were set above..

jlaine:
This looks like it needs polishing:

- the UX looks odd: you should only offer this selection if video is selected, and keep it grouped with other video-related options
- the camera selection shouldn't be mandatory, it worked fine without it for most cases
- selecting the camera should not clear the resolution settings

jlaine:
Closing in favour of #871 which also make it possible to select the audio device

PR #907, opened 2023-07-13
Move VP8 Codec Parameters to Constants

MorlaxAR:
Simple change to stop using literals in some key VP8 codec parameter settings.

Currently there is no way to specify the codec parameters. Defining them as constants allows them to be easily modified.
Default values were left as is, and I have verified that the codec works when modifying the exposed parameters.


jlaine:
Constants are supposed to be.. constant? I have a feeling you plan to monkey patch aiortc, which is not future-proof.

Do the WebRTC standards provide a mechanism to customise codec settings? If so I'd rather we adopt such an API

Issue #828, opened 2023-02-10
DTLS Bio read buffer is to small and not compatible with ecdsa-with-SHA256 key/cert with CA


zucher:
With a Janus DTLS certificate with 4096 size, DTLS  handshake goes in timeout, the transsaction never succeed.

a related issue was solved few year ago on janus side: https://github.com/meetecho/janus-gateway/issues/252

after investiguation it appears that the ssl bio_read buffer is too small to read the full certificate.

see https://github.com/aiortc/aiortc/blob/d30b6f75d4457204b1ab336106a248d535be053a/src/aiortc/rtcdtlstransport.py#L630

I do some test ad I suggest to increase the bio_read buffer size
from:
```python
try:
            data = self.ssl.bio_read(1500)
        except SSL.Error:
            data = b""
...
``` 
to a bigger value
```python
try:
            data = self.ssl.bio_read(8192)
        except SSL.Error:
            data = b""
...
``` 
It will solves also this closed issue: https://github.com/aiortc/aiortc/issues/346

an alternative but not tested, could be to read in a loop all the bio_read data, and send the full buffer after at https://github.com/aiortc/aiortc/blob/d30b6f75d4457204b1ab336106a248d535be053a/src/aiortc/rtcdtlstransport.py#L634

dfybc:
hi，
I increase the bio_read buffer size from 1500 to 8192, but error still occured.
i send 16kb data every 0.1 second through the data channel。
error happened as following:
2023-02-15 13:59:06.453708 - aiortc.rtcdtlstransport - WARNING - RTCDtlsTransport(client) Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/aiortc/rtcdtlstransport.py", line 460, in __run
    await self._recv_next()
  File "/usr/local/lib/python3.7/dist-packages/aiortc/rtcdtlstransport.py", line 545, in _recv_next
    await self._write_ssl()
  File "/usr/local/lib/python3.7/dist-packages/aiortc/rtcdtlstransport.py", line 630, in _write_ssl
    data = self.ssl.bio_read(8192)
  File "/usr/local/lib/python3.7/dist-packages/OpenSSL/SSL.py", line 2029, in bio_read
    self._handle_bio_errors(self._from_ssl, result)
  File "/usr/local/lib/python3.7/dist-packages/OpenSSL/SSL.py", line 2004, in _handle_bio_errors
    raise ValueError("unknown bio failure")
[...]

PR #921, opened 2023-08-28
Increase DTLS bios buffer size for large x509

zucher:
Solves the issue https://github.com/aiortc/aiortc/issues/828 for large X509 certificate such as RSA 4096 in DTLS exchanges

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -616,7 +616,7 @@ class RTCDtlsTransport(AsyncIOEventEmitter):
         Flush outgoing data which OpenSSL put in our BIO to the transport.
         """
         try:
-            data = self.ssl.bio_read(1500)
+            data = self.ssl.bio_read(8192)
         except SSL.Error:
             data = b""
         if data:


lgrahl:
Are you sure it solves the issue portably? I would expect this code to just send a large UDP datagram which is undesirable. Can you verify what happens with a Wireshark trace?

zucher:
In complement, ssl provide a bio_pending function dedicated to wait leaving packets:
https://www.openssl.org/docs/man3.0/man3/BIO_pending.html and this function confirm the capability to receive multiple UDP packets if the content is larger than an MTU

jlaine:
Hi @zucher, I'm a little concerned that raising the 1500 bytes arbitrary limit to another one won't fix fix the issue. Could we please have a unit test which demonstrates we are actually able to send / receive larger certificates?

zucher:
Hi @jlaine, unit test are not so easy to produce, are the pcap files not sufficient?

zucher:
And in complement, openssl documentation too?

jlaine:
> Hi @jlaine, unit test are not so easy to produce, are the pcap files not sufficient?

Unfortunately no they are not. I have put in a lot of effort to achieve 100% test coverage to ensure the code behaves as it should and new contributions need to do the same.

I am of course willing to assist you in putting together the tests, but will not be able to write them for you to exercise a condition I have not yet encountered. Could you shed some light on what makes the certificates "large"? A first step would be to have aiortc generate such certificates so we can capture the failure case.

lgrahl:
@zucher I think all you need to do is create a large RSA certificate. The tests then could ensure that the MTU isn't exceeded / the certificate is fragmented across multiple UDP datagrams.

jlaine:
> @zucher I think all you need to do is create a large RSA certificate. The tests then could ensure that the MTU isn't exceeded / the certificate is fragmented across multiple UDP datagrams (**not** using IP fragmentation).

As my two parties are going to be `aiortc` I'd really like to understand why this line in the proposed PR is the one being changed, as this is very much for *sending* data:

https://github.com/aiortc/aiortc/pull/921/files#diff-4637ab9c11f102ccb8bcc021ce74a3460d8c84d18f7d6bf67a02d618debee1f8R619

zucher:
@jlaine ,you are right, had this issue as client due to a connection with a Janus server with such certificate, but if such certificate is use on aiortc side, probably the bio_write call might be also involved. I try to undersand the test suite to validate all cases.

fippo:
The actual solution is more complicated, see https://github.com/meetecho/janus-gateway/pull/254
Avoiding RSA certificates is much easier

jlaine:
OK I'm going to close this PR, as it doesn't seem to actually solve the problem.

PR #991, opened 2023-11-27
Use set_source_pkgconfig instead of set_source

Sean-Der:
I don't fully understanding how Python does FFI, but I don't understand how it would work before.

Now it uses pkg-config and will pull the header locations properly.

### DIFF
--- a/src/_cffi_src/build_opus.py
+++ b/src/_cffi_src/build_opus.py
@@ -1,14 +1,13 @@
-from cffi import FFI
+import cffi
 
-ffibuilder = FFI()
+module_name = "aiortc.codecs._opus"
 
-ffibuilder.set_source(
-    "aiortc.codecs._opus",
-    """
-#include <opus/opus.h>
-    """,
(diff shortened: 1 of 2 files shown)


jlaine:
Hi @Sean-Der thanks for the PR! Let's let CI run to see how it turns out. I'm a little doubtful this will work on Windows?

Sean-Der:
@jlaine thanks!

Should I put this behind a platform check? I don’t understand how it works today on Mac/Linux. If those headers aren’t in /usr/include how does it find it?

I needed this change to build on NixOS and my Mac.

jlaine:
@Sean-Der I've taken a slightly different approach of trying `pkg-config` first and falling back to our previous logic if that fails. Could you let me know if it works for you?

jlaine:
Ping @Sean-Der ?

jlaine:
I'm closing this, it's no longer relevant since we don't link against any libraries anymore.

PR #1011, opened 2023-12-05
Update DTLS cipher list

### LINKED ISSUE #956: Proposal for an API to set DTLS cipher suites
Hi,

Would you consider adding an API to set the DTLS cipher suite?

Let me explain the background behind this proposal.
I was trying to create an application to send video over [WHIP](https://www.ietf.org/archive/id/draft-ietf-wish-whip-01.html) between [OBS](https://github.com/obsproject/obs-studio/) and aiortc.
When I tested the operation, I found that the DTLS handshake was failing.
After examining the code in OBS, [libdatachannel](https://github.com/paullouisageneau/libdatachannel), and [Mbed TLS](https://github.com/Mbed-TLS/mbedtls), I found that the cause was that [Mbed TLS does not support the Client Hello fragmentation.](https://github.com/Mbed-TLS/mbedtls/blob/v3.5.0/library/ssl_tls12_server.c#L1083)
As a side note, OBS uses libdatachannel as its WebRTC library, which is built with a configuration that uses Mbed TLS.

The aiortc is [configured](https://github.com/aiortc/aiortc/blob/main/src/aiortc/rtcdtlstransport.py#L155) with `HIGH:!CAMELLIA:!aNULL` as the cipher suite and the list size was 56 in my environment.
I found that by choosing the cipher suite to set, I could reduce the size of the list and avoid the fragmentation of Client Hello.

The code implementing the API to set the cipher suite has been committed to [my forked repository](https://github.com/aiortc/aiortc/compare/main...kadoshita:aiortc:add-set-cipher-of-dtls) and tests have been written.
If you agree with this proposal, I will send you a Pull Request.

I understand that this is a feature that does not exist in the standard WebRTC API, but I believe that this API would increase aiortc interoperability.

kadoshita:
Fix #956 

The current DTLS cipher suite setting is `HIGH:!CAMELLIA:!aNULL` , in which case the Client Hello fragmentation occurs.
Therefore, if a library that does not support the Client Hello fragmentation, such as Mbed TLS, is used, the handshake will fail.

This problem can be solved by aligning the cipher suite to the browser and reducing the number of cipher suites.

|before|after|
|:--:|:--:|
|before|after|

In my environment, the number of cipher suites sent by Client Hello is 12, as shown below, and I checked that fragmentation does not occur.

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -191,7 +191,9 @@ class RTCCertificate:
         )
         ctx.use_certificate(self._cert)
         ctx.use_privatekey(self._key)
-        ctx.set_cipher_list(b"HIGH:!CAMELLIA:!aNULL")
+        ctx.set_cipher_list(
+            b"ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA"
+        )
         ctx.set_tlsext_use_srtp(b":".join(x.openssl_profile for x in srtp_profiles))
 
         return ctx


lgrahl:
[RFC 8827 states that](https://datatracker.ietf.org/doc/html/rfc8827#section-6.5) only `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` with P-256 is mandatory and this aligns with what I've seen in the wild. Considering this and that aiortc does not like large RSA certificates (#828), I think we should also exclude all RSA cipher suites.

kadoshita:
Thank you.

I added `:!aRSA` and changed it to exclude RSA.
The cipher suites are as follows after changed.

```
Cipher Suites (5 suites)
    Cipher Suite: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 (0xc02b)
    Cipher Suite: TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 (0xcca9)
    Cipher Suite: TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA (0xc009)
    Cipher Suite: TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA (0xc00a)
    Cipher Suite: TLS_EMPTY_RENEGOTIATION_INFO_SCSV (0x00ff)
```

jlaine:
Hi! I'm a bit confused by the string we are passing: since we're explicitly settings some cipher suites, why do we also need to exclude RSA?

kadoshita:
For example, AES128-SHA uses RSA as the key exchange algorithm and authentication algorithm.
The Google Chrome cipher suite included AES128-SHA, so we followed it.
However, I received a comment that cipher suites that use RSA should be excluded, so I updated it to exclude RSA.

Instead of excluding RSA, should we exclude cipher suites that use RSA as a key exchange or authentication algorithm?

src/aiortc/rtcdtlstransport.py
```diff
@@ -152,7 +152,7 @@ def _create_ssl_context(self) -> SSL.Context:
         )
         ctx.use_certificate(self._cert)
         ctx.use_privatekey(self._key)
-        ctx.set_cipher_list(b"HIGH:!CAMELLIA:!aNULL")
+        ctx.set_cipher_list(b"ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES128-SHA:AES256-SHA:!aRSA")
```
jlaine, line 155:
```suggestion
        ctx.set_cipher_list(b"ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA:AES128-GCM-SHA256:AES128-SHA:AES256-SHA")
```

jlaine:
My point was: why declare RSA-based cipher suites then disable them with `!aRSA`, why not simply pull them from the list?

kadoshita:
Sorry.
I think it would be better to exclude RSA-based cipher suites from the list, as you suggested.

jlaine:
I've squashed down your commits, rebased on top of `main` and made the commit a bit more descriptive.

@fippo @lgrahl could you please comment on whether this cipher list is OK so we can get this merged?

jlaine (approved):

kadoshita:
@jlaine @fippo @lgrahl Thank you for your reviewing!

PR #1054, opened 2024-02-24
Include musllinux builds

TECH7Fox:
Removes the filter to include `musllinux` builds for Alpine, which is very popular for containers, and more. In my case I need it specifically for Home Assistant. [Tested it](https://github.com/TECH7Fox/aiortc/actions/runs/8027902526) and the CI runs fine and includes the additional `musllinux` wheels.

Was there a reason to originally exclude it?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -124,7 +124,6 @@ jobs:
           CIBW_BEFORE_BUILD_WINDOWS: python scripts\fetch-vendor.py C:\cibw\vendor
           CIBW_ENVIRONMENT: CFLAGS=-I/tmp/vendor/include LDFLAGS=-L/tmp/vendor/lib
           CIBW_ENVIRONMENT_WINDOWS: INCLUDE=C:\\cibw\\vendor\\include LIB=C:\\cibw\\vendor\\lib
-          CIBW_SKIP: '*-musllinux*'
         run: |
           pip install cibuildwheel
           cibuildwheel --output-dir dist


TECH7Fox:
@jlaine could you take a look at this? Would be very useful since we can't deploy to HA without this.

jlaine:
While I'm not opposed to this as such, this is not a trivial effort. The first step is going to be to make sure there are `musllinux` wheels for all binary dependencies:

- [ ] https://pypi.org/project/av/  This one is going to be *hard*. It involves building FFmpeg from source.
- [x] https://pypi.org/project/cryptography/
- [x] https://pypi.org/project/google-crc32c/
- [ ] https://pypi.org/project/pylibsrtp/ 

It almost certainly involves producing binary builds of `opus` and `vpx` for musllinux:

https://github.com/aiortc/aiortc-codecs

The build system also needs to be updated to fetch the right version of these builds and put them in their own directory.

jlaine:
The linting error will go away if you rebase on top of `main`, which should allow you to view the failures with the wheels.

TECH7Fox:
Thank you for the message. For now I got aiortc successfully building in https://github.com/home-assistant/wheels-custom-integrations/pull/684

So we can use it in HA. But this would still be helpful for others.

I will try to get this working when I have some time.

jlaine:
Hi! Do you still plan to complete this PR?

jlaine:
I'm closing this, as `aiortc` is now pure Python.

PR #1055, opened 2024-02-26
enable|disable RTCRtpSender and RTCRtpReceiver

### LINKED ISSUE #1053: Setting RTCRtpTransceiver.direction to `inactive` does not have any effect
We have a transceiver that was generated out of a `pc.addTransceiver(track)`, meaning the transceiver has only a receiver.

When setting the transceiver direction to `inactive`, while we keep sending RTP to this endpoint we expect it to just ignore the traffic, but `RTCRtpReceiver.__handle_rtp_packet()` simply processes the traffic normally.

My expectation is that the receiver should ignore the incoming RTP packet. In our case we look for the receiver not to process anything, that's why we set the direction to 'inactive'.

jmillan:
Fixes #1053 

Enable or disable them based on the negotiated direction.

This avoids RTCRtpSender sending RTP and RTCRtpReceiver processing incoming RTP if disabled.

### DIFF
--- a/src/aiortc/rtcrtpreceiver.py
+++ b/src/aiortc/rtcrtpreceiver.py
@@ -263,6 +263,7 @@ class RTCRtpReceiver:
         if transport.state == "closed":
             raise InvalidStateError
 
+        self._enabled = True
         self.__active_ssrc: Dict[int, datetime.datetime] = {}
         self.__codecs: Dict[int, RTCRtpCodecParameters] = {}
         self.__decoder_queue: queue.Queue = queue.Queue()
@@ -449,6 +450,10 @@ class RTCRtpReceiver:
         Handle an incoming RTP packet.
         """
         self.__log_debug("< %s", packet)
+
+        # If the receiver is disabled, discard the packet.
+        if not self._enabled:
+            return
 
         # feed bitrate estimator
         if self.__remote_bitrate_estimator is not None:
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -78,6 +78,7 @@ class RTCRtpSender:
         self._rtx_ssrc = random32()
         # FIXME: how should this be initialised?
         self._stream_id = str(uuid.uuid4())
+        self._enabled = True
         self.__encoder: Optional[Encoder] = None
         self.__force_keyframe = False
         self.__loop = asyncio.get_event_loop()
@@ -268,6 +269,13 @@ class RTCRtpSender:
     async def _next_encoded_frame(self, codec: RTCRtpCodecParameters):
         # get [Frame|Packet]
         data = await self.__track.recv()
+
+        # If the sender is disabled, drop the frame instead of encoding it.
+        # We still want to read from the track in order to avoid frames
+        # accumulating in memory.
+        if not self._enabled:
+            return None
+
         audio_level = None
 
         if self.__encoder is None:
@@ -325,7 +333,12 @@ class RTCRtpSender:
                     await asyncio.sleep(0.02)
                     continue
 
+                # Fetch the next encoded frame. This can be `None` if the sender
+                # is disabled, in which case we just continue the loop.
                 enc_frame = await self._next_encoded_frame(codec)
+                if enc_frame is None:
+                    continue
+
                 timestamp = uint32_add(timestamp_origin, enc_frame.timestamp)
 
                 for i, payload in enumerate(enc_frame.payloads):
--- a/src/aiortc/rtcrtptransceiver.py
+++ b/src/aiortc/rtcrtptransceiver.py
@@ -29,6 +29,7 @@ class RTCRtpTransceiver:
         sender: RTCRtpSender,
         direction: str = "sendrecv",
     ):
+        self.__currentDirection: Optional[str] = None
         self.__direction = direction
         self.__kind = kind
         self.__mid: Optional[str] = None
@@ -37,7 +38,6 @@ class RTCRtpTransceiver:
         self.__sender = sender
         self.__stopped = False
 
-        self._currentDirection: Optional[str] = None
         self._offerDirection: Optional[str] = None
         self._preferred_codecs: List[RTCRtpCodecCapability] = []
         self._transport: RTCDtlsTransport = None
@@ -54,7 +54,7 @@ class RTCRtpTransceiver:
 
         One of `'sendrecv'`, `'sendonly'`, `'recvonly'`, `'inactive'` or `None`.
         """
-        return self._currentDirection
+        return self.__currentDirection
 
     @property
     def direction(self) -> str:
@@ -129,6 +129,22 @@ class RTCRtpTransceiver:
         await self.__receiver.stop()
         await self.__sender.stop()
         self.__stopped = True
+
+    def _setCurrentDirection(self, direction: str) -> None:
+        self.__currentDirection = direction
+
+        if direction == "sendrecv":
+            self.__sender._enabled = True
+            self.__receiver._enabled = True
+        elif direction == "sendonly":
+            self.__sender._enabled = True
+            self.__receiver._enabled = False
+        elif direction == "recvonly":
+            self.__sender._enabled = False
+            self.__receiver._enabled = True
+        elif direction == "inactive":
+            self.__sender._enabled = False
+            self.__receiver._enabled = False
 
     def _set_mid(self, mid: str) -> None:
         self.__mid = mid
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -789,7 +789,7 @@ class RTCPeerConnection(AsyncIOEventEmitter):
         # configure direction
         for t in self.__transceivers:
             if description.type in ["answer", "pranswer"]:
-                t._currentDirection = and_direction(t.direction, t._offerDirection)
+                t._setCurrentDirection(and_direction(t.direction, t._offerDirection))
 
         # gather candidates
         await self.__gather()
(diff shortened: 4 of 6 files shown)


jmillan:
I've used an Intel laptop since with the M1 I was unable to run the tests.

jmillan:
I'm willing to work on the tests but I'm still unable to build the package in my M1, which is where I have my dev setup.

jmillan:
@jlaine would you please take a couple of minutes to check this PR?

It's straight forward and quite clean.

The bug this PR fixes is preventing us from using aiortc for load testing a WebRTC environent.

ibc:
Just a friendly ping because we are in a situation in which we already had to fork `aiortc` fo have this issue fixed.

jlaine:
Hi! I'm currently away but will definitely take a look at this PR as soon as possible.

One thing which is not clear to me is what the expected behaviour is for a disabled *receiver*: return silent/blank frames or nothing at all. Could you please provide some links to specs?

By the way two weeks seems like a low bar for a fork :)
Any interest in co-maintaining aiortc?

src/aiortc/rtcrtptransceiver.py
```diff
@@ -54,7 +54,24 @@ def currentDirection(self) -> Optional[str]:
 
         One of `'sendrecv'`, `'sendonly'`, `'recvonly'`, `'inactive'` or `None`.
         """
-        return self._currentDirection
+        return self.__currentDirection
+
+    @currentDirection.setter
+    def currentDirection(self, direction: str) -> None:
```
jlaine, line 60: This looks wrong, `currentDirection` is supposed to be read-only:

https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver

ibc:
> One thing which is not clear to me is what the expected behaviour is for a disabled _receiver_: return silent/blank frames or nothing at all. Could you please provide some links to specs?

Perhaps you mean "disabled _sender_"? If so it shouldn't send anything. In WebRTC there is no need to artificially send RTP all the time (silence or empty video frames while disabled) since there is already mandatory ICE consent mechanism running so there is proper keepalive system already. I don't know if I could be able to find links to spec about this.


> By the way two weeks seems like a low bar for a fork :) Any interest in co-maintaining aiortc?

Last thing we want is to maintain a fork. Sorry for the rush. Not sure if we have bandwidth now to maintain yet another project. We'll keep contributing to it as much as we can but we are not Python guys at all.

jmillan:
> One thing which is not clear to me is what the expected behaviour is for a disabled receiver: return silent/blank frames or nothing at all. Could you please provide some links to specs?

Setting the direction of a `RtcRtpTransceiver` will take effect in future calls of `createOffer` or `createAnswer` in order to mark the corresponding media description as sendrecv, sendonly, recvonly or inactive. So no RTP sending/procesing is expected when disabling sender or receiver respectively.

src/aiortc/rtcrtptransceiver.py
jlaine: I would probably go for a `_setCurrentDirection` method to make it clear this is not public API

src/aiortc/rtcrtpreceiver.py
```diff
@@ -398,6 +399,14 @@ async def receive(self, parameters: RTCRtpReceiveParameters) -> None:
...
         self.__transport = transport
 
+    @property
+    def enabled(self) -> bool:
+        return self.__enabled
+
+    @enabled.setter
+    def enabled(self, enabled: bool) -> None:
+        self.__enabled = enabled
+
```
jlaine, line 409: I don't see any such property in the corresponding JS interface, so this seems odd?

src/aiortc/rtcrtpsender.py
```diff
@@ -175,6 +176,14 @@ def replaceTrack(self, track: Optional[MediaStreamTrack]) -> None:
         else:
             self._track_id = str(uuid.uuid4())
 
+    @property
+    def enabled(self) -> bool:
+        return self.__enabled
+
+    @enabled.setter
+    def enabled(self, enabled: bool) -> None:
+        self.__enabled = enabled
```
jlaine, line 185: Same as for `RTCRtpReceiver`, this property does not exist on in the JS interface. If it's just for our internal use, maybe just use a `_enabled` (single underscore).

src/aiortc/rtcrtpsender.py
```diff
@@ -321,7 +330,7 @@ async def _run_rtp(self, codec: RTCRtpCodecParameters) -> None:
         timestamp_origin = random32()
         try:
             while True:
-                if not self.__track:
+                if not self.__track or not self.__enabled:
```
jlaine, line 333: OK, now here things start getting worrying. This means that if if `enabled` is false, we're not consuming media from the track, which completely changes the media flow. For example if your media source is a webcam, and you disable sending you're going to blow up your RAM because frames are going to accumulate, or am I wrong?

jlaine:
On the whole, I'm not against this PR, but it feels like one of those cases where we are going to run against limits of how aiortc's "media pipeline" was designed. I am starting to feel like the media pipeline just doesn't match how other frameworks implement it, but I really don't have a good view of what I *should* have implemented.

src/aiortc/rtcrtpreceiver.py
jmillan: True, this could be `_enabled()` to indicate it's private.

src/aiortc/rtcrtpsender.py
jmillan: Currently you'll find a very similar situation if you negotiate the reception of RTP and don't do anything with it. Memory will grow indefinitely as you don't consume the incoming media.

IMO if the user disables sending RTP then she must also stop the track, in order to avoid accumulating frames.

src/aiortc/rtcrtptransceiver.py
jmillan: I renamed the setter from `currentDirection` to `_currentDirection`

src/aiortc/rtcrtpsender.py
ibc: I'm a bit confused about this topic. In browser I can do getUserMedia(), obtain tracks for mic and webcam and not use them anywhere, and they are not gonna produce any leak. This is real-time world. If something is producing audio units of video frames and nobody is consuming them then they should just be discarded.

jmillan: The case described by @jlaine corresponds to:
* You create a transceiver with a sender track.
* You don't send any traffic because sender is disabled.

In this case the frames in the track are not being consumed hence the track is accumulating them.

> If something is producing audio units of video frames and nobody is consuming them then they should just be discarded.

Yes, aiortc could, in case of being disabled, retrieve the frames from the track as if it was going to be sent, but simply discard it.

jmillan: This is not a concern anymore. Handled here https://github.com/aiortc/aiortc/pull/1055/commits/8faf5318ccb4a688813ae17503a6203761f893d7

jlaine:
You can ignore the linter error for now, I'm fixing this in #1058

jmillan:
@jlaine, all the concerns are fixed now :-)

src/aiortc/rtcrtpsender.py
```diff
@@ -175,6 +176,14 @@ def replaceTrack(self, track: Optional[MediaStreamTrack]) -> None:
...
             self._track_id = str(uuid.uuid4())
 
+    @property
+    def _enabled(self) -> bool:
+        return self.__enabled
+
+    @_enabled.setter
+    def _enabled(self, enabled: bool) -> None:
+        self.__enabled = enabled
+
```
jlaine, line 186: Sorry I should have been clearer here: I don't think we need getter / setters at all, just initial the property as `self._enabled = True` in the constructor (`__init__`).

src/aiortc/rtcrtpreceiver.py
```diff
@@ -398,6 +399,14 @@ async def receive(self, parameters: RTCRtpReceiveParameters) -> None:
...
         self.__transport = transport
 
+    @property
+    def _enabled(self) -> bool:
+        return self.__enabled
+
+    @_enabled.setter
+    def _enabled(self, enabled: bool) -> None:
+        self.__enabled = enabled
+
```
jlaine, line 409: Same as for `RTCRtpSender`, just initialise `self._enabled = True` in the constructor, we don't need a getter / setter.

src/aiortc/rtcrtpsender.py
```diff
@@ -325,6 +334,11 @@ async def _run_rtp(self, codec: RTCRtpCodecParameters) -> None:
                     await asyncio.sleep(0.02)
                     continue
 
+                # retrieve the next frame from the track and discard it
+                if not self.__enabled:
+                    await self._next_encoded_frame(codec)
+                    continue
+
                 enc_frame = await self._next_encoded_frame(codec)
```
jlaine, line 342:
```suggestion
                enc_frame = await self._next_encoded_frame(codec)
                
                # If the sender is disabled, just discard the frame.
                if not self.__enabled:
                    continue
                
```

jlaine:
OK the code looks like something I'm happy to merge, do you mind if I take over the branch to edit some comments?

jmillan:
> OK the code looks like something I'm happy to merge, do you mind if I take over the branch to edit some comments?

Sure, please.

jlaine:
OK, I've made the changes I wanted:

- I've added some comments.
- I changed the `_currentDirection` setter to a method, because I don't understand why `@currentDirection.setter` (no leading underscore) worked
- I introduced a slight optimisation: if the sender is disabled, don't run frames through the encoder.

Let's see how CI feels about the changes!

jlaine:
@jmillan @ibc I've merged the PR and cut a new release, so you should soon have wheels on PyPI !

PR #1090, opened 2024-05-04
Implement handling of non-actpass setup in offers

### LINKED ISSUE #1087: aiortc rejecting offer with setup=passive against RFC guidance
Repro steps
1. Have an SDP that has `a=setup:passive` in it
2. Create `RTCSessionDescription` with this sdp and `type="offer"`
3. Call `pc.setRemoteDescription` with above session description

Expected
1. This succeeds. See [RFC spec](https://datatracker.ietf.org/doc/html/rfc8842#section-5.3-6) that states receiver should be prepared for both `active` and `passive` as well.

> Even though an offerer is required to insert an "SDP" setup attribute with an "actpass" attribute value in initial offers ([Section 5.2](https://datatracker.ietf.org/doc/html/rfc8842#sec-oa-offer)) and subsequent offers ([Section 5.5](https://datatracker.ietf.org/doc/html/rfc8842#sec-oa-mod)), the **_answerer MUST be able to receive initial and subsequent offers with other attribute values_**, in order to be backward compatible with older implementations that might insert other attribute values in initial and subsequent offers.

Actual
1. An exception is raised with message "DTLS setup attribute must be 'actpass' for an offer" from [__validate_description method](https://github.com/aiortc/aiortc/blob/e9c13eab915ddc27f365356ed9ded0585b9a1bb7/src/aiortc/rtcpeerconnection.py#L1277)

PS: I see this got discussed a bit in https://github.com/aiortc/aiortc/issues/267

fippo:
RFC 8842 updated RFC 5763 to allow a differt setup attribute than
  actpass
in offers which allows for more control of the DTLS role.

See https://issues.webrtc.org/issues/42223106 for additional details

Fixes #1087

### DIFF
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -928,6 +928,8 @@ class RTCPeerConnection(AsyncIOEventEmitter):
                     iceTransport._role_set = True
 
                 # set DTLS role
+                if description.type == "offer" and media.dtls.role == "client":
+                    dtlsTransport._set_role(role="server")
                 if description.type == "answer":
                     dtlsTransport._set_role(
                         role="server" if media.dtls.role == "client" else "client"
@@ -1273,8 +1275,6 @@ class RTCPeerConnection(AsyncIOEventEmitter):
                 raise ValueError("ICE username fragment or password is missing")
 
             # check DTLS role is allowed
-            if description.type == "offer" and media.dtls.role != "auto":
-                raise ValueError("DTLS setup attribute must be 'actpass' for an offer")
             if description.type in ["answer", "pranswer"] and media.dtls.role not in [
                 "client",
                 "server",
--- a/tests/test_rtcpeerconnection.py
+++ b/tests/test_rtcpeerconnection.py
@@ -4686,30 +4686,6 @@ a=rtpmap:0 PCMU/8000
         await pc2.close()
 
     @asynctest
-    async def test_setRemoteDescription_with_invalid_dtls_setup_for_offer(self):
-        pc1 = RTCPeerConnection()
-        pc2 = RTCPeerConnection()
-
-        # apply offer
-        pc1.addTrack(AudioStreamTrack())
-        offer = await pc1.createOffer()
-        await pc1.setLocalDescription(offer)
-        mangled = RTCSessionDescription(
-            sdp=pc1.localDescription.sdp.replace("a=setup:actpass", "a=setup:active"),
-            type=pc1.localDescription.type,
-        )
-        with self.assertRaises(ValueError) as cm:
-            await pc2.setRemoteDescription(mangled)
-        self.assertEqual(
-            str(cm.exception),
-            "DTLS setup attribute must be 'actpass' for an offer",
-        )
-
-        # close
-        await pc1.close()
-        await pc2.close()
-
-    @asynctest
     async def test_setRemoteDescription_with_invalid_dtls_setup_for_answer(self):
         pc1 = RTCPeerConnection()
         pc2 = RTCPeerConnection()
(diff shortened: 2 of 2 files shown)


src/aiortc/rtcpeerconnection.py
```diff
@@ -928,6 +928,8 @@ async def setRemoteDescription(
                     iceTransport._role_set = True
 
                 # set DTLS role
+                if description.type == "offer" and media.dtls.role == "client":
+                    dtlsTransport._set_role(role="server")
                 if description.type == "answer":
                     dtlsTransport._set_role(
                         role="server" if media.dtls.role == "client" else "client"
```
lgrahl, line 935: Btw what role do we end up with when none of these conditionals match?

lgrahl (commented):
Looks correct to me.

src/aiortc/rtcpeerconnection.py
fippo: `auto` which I *hope* means "determined by who wins ICE"

src/aiortc/rtcpeerconnection.py
```diff
@@ -1272,9 +1274,6 @@ def __validate_description(
             if not media.ice.usernameFragment or not media.ice.password:
                 raise ValueError("ICE username fragment or password is missing")
 
-            # check DTLS role is allowed
-            if description.type == "offer" and media.dtls.role != "auto":
```
Ragsboss, line 1276: Do we need to also remove raising error in `__validate_description` when peer's role in peer's offer is not `actpass`? https://github.com/aiortc/aiortc/blob/e9c13eab915ddc27f365356ed9ded0585b9a1bb7/src/aiortc/rtcpeerconnection.py#L1277

jlaine:
Hi @fippo ! Thanks for looking into this.

I must admit I'm confused by the wording in RFC 8842, it stills say that offers should use `actpass`.. but allow other values for *backwards* compatibility?

https://datatracker.ietf.org/doc/html/rfc8842#name-generating-the-initial-sdp-

fippo:
Backward compatible in the sense "if you don't know the other side supports this keep using actpass but if you know it does (or assume)... then go ahead". For this to work we need more implementation (like aiortc) to support the new style ;-)

src/aiortc/rtcpeerconnection.py
jlaine: I'm not sure I understand why we are only handling `media.dtls.role == "client"`?

fippo: This is the handling of offers so the existing code assumes actpass. Lets see if I can prove that by adding more tests because who does not like more tests ;-)

jlaine:
Rebasing on top of `main` will fix the error building wheels on macOs / x86_64.

src/aiortc/rtcpeerconnection.py
fippo: As I suspected... the current code assumes role=auto in SRD. Which is not quite correct since auto means "determined by DTLS" whereas when we parse actpass we should be setting this to "client" explicitly. This is currently done in createAnswer which is a bit confusing but ok.

fippo:
> Rebasing on top of `main` will fix the error building wheels on macOs / x86_64.

That fixed one job but another started failing 😂

src/aiortc/rtcpeerconnection.py
```diff
@@ -1272,9 +1274,6 @@ def __validate_description(
             if not media.ice.usernameFragment or not media.ice.password:
                 raise ValueError("ICE username fragment or password is missing")
 
-            # check DTLS role is allowed
```
jlaine, line 1275: I think we still want this comment

jlaine:
Thanks so much @fippo!

Issue #792, opened 2022-11-02
How to use AioRTC with just Async / Loop / Threading calls.


R0NAM1:
Before filing an issue please verify the following:

* Check whether there is already an existing issue for the same topic.
_**(Not from what I could find)**_
* Ensure you are actually reporting an issue related to ``aiortc``. The goal
  of the issue tracker is not to provide general guidance about WebRTC or free
  debugging of your code.
_**(Yes, it's explicitly related on how to call AioRTC properly when not using AioHTTP,)**_
* Clearly state whether the issue you are reporting can be reproduced with one
  of the examples provided with ``aiortc`` *without any changes*.
_**(I was able to take the SERVER example and convert it to Flask, the only difference seems to be how it's called.)**_
* Be considerate to the maintainers. ``aiortc`` is provided on a best-effort,
  there is no guarantee your issue will be addressed.

I've been attempting to use AioRTC with Flask to stream video to HTML5 Clients, so far everything works correctly EXCEPT the server refuses to accept the User Bind Request from the client (Verified via Wireshark). I believe the issue is with how I'm executing some of the async code, it seems examples using AioHTTP can do this no problem, it takes care of it for them, but for Flask (Which is what I'm using), you need to make the async calls manually.

In the logs, running the loop until completion the final debug messages for ICE are:
REQUEST
RESPONSE

But if after I put loop.run_forever, then I get the following:
REQUEST
RESPONSE
WAITING
IN_PROGRESS
SUCCEEDED

The issue with running it forever is that I don't get my SDP return for the client, so even though the server can now actually get ICE candidates, now the client can't. 

How can this be addressed?

```
async def webRtcStart():
    loop = asyncio.get_event_loop()
    
    global rtspCredString

    player = MediaPlayer(rtspCredString)    
    params = ast.literal_eval((request.data).decode("UTF-8"))
    
    # print(params)

    offer = RTCSessionDescription(sdp=params.get("sdp"), type=params.get("type"))
    webRtcPeer = RTCPeerConnection(configuration=RTCConfiguration(
    iceServers=[RTCIceServer(
        urls=['stun:stun.internal.my.domain'])]))

    logger = logging.getLogger("webRtcPeer")

    logging.basicConfig(level=logging.DEBUG)

    



    if (player.video):
        webRtcPeer.addTrack(player.video)

    await webRtcPeer.setRemoteDescription(offer)

    print(offer)


    answer = await webRtcPeer.createAnswer()

[...]

PR #1103, opened 2024-05-28
Change RTCP Sender to use asyncio.Event() instead of a task cancelled…

R0NAM1:
… exception as that is unreliable after hours of usage, original task.cancel() kept as it is still needed to close it, new while just makes sure it can close.

I've been creating an NVR software based around Flask and AioRTC, I have observed through rigorous testing that after extensive time alive (5+ hours) rtcp_run does not get the task.cancel() signal, and so the loop continues forever never stopping, meaning I cannot pc.close() after long sessions. 
(Which will happen with an NVR system).

This pull request implements a fix for this by replacing the Cancel signal with an asyncio.Event() object that I have observed reliably delivering a signal and allowing rtcp_run to die. rtc_run has no issues and quits the second the peer closes the tab.

You can find the relevant thread here:
https://github.com/aiortc/aiortc/issues/792

(I'm currently testing this fix for the final time with all my other session quitting issues fixed, so I'll have an answer tomorrow morning.)

My project for reference (Relevant code not present, have not committed yet.):
https://github.com/R0NAM1/zemond

### DIFF
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -90,6 +90,7 @@ class RTCRtpSender:
         self.__rtp_history: Dict[int, RtpPacket] = {}
         self.__rtcp_exited = asyncio.Event()
         self.__rtcp_started = asyncio.Event()
+        self.__rtcp_quitting = asyncio.Event()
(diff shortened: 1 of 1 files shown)


R0NAM1:
Did not fix issue, doing further troubleshooting.

PR #1104, opened 2024-05-29
Change RTCP Sender to use asyncio.Event() instead of a task cancelled…

R0NAM1:
… exception as that is unreliable after hours of usage, original task.cancel() kept as it is still needed to close it, new while just makes sure it can close. Also wrap _send_rtcp in wait_for as it fails after a few hours and awaits forever.

Amendment of https://github.com/aiortc/aiortc/pull/1103. 

I've noticed after prolonged usage (1 hour+) the _send_rtcp function will await forever and thus cannot be cancelled when closing. Wrapping this in a wait_for function will cause code to continue when this hangs. This also shows that the Task Cancelled event does not reliably get to the loop so it can break sometimes, so I replaced this with a regular asyncio.Event() and it seems to work again.

Also check
https://github.com/aiortc/aiortc/issues/792

### DIFF
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -90,6 +90,7 @@ class RTCRtpSender:
         self.__rtp_history: Dict[int, RtpPacket] = {}
         self.__rtcp_exited = asyncio.Event()
         self.__rtcp_started = asyncio.Event()
+        self.__rtcp_quitting = asyncio.Event()
         self.__rtcp_task: Optional[asyncio.Future[None]] = None
         self.__rtx_payload_type: Optional[int] = None
         self.__rtx_sequence_number = random16()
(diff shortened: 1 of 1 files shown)


jlaine:
This is going to require a bit more explanation, I don't understand what we're doing here. Why does the duration for which the code run affect delivery of the cancellation?

R0NAM1:
> This is going to require a bit more explanation, I don't understand what we're doing here. Why does the duration for which the code run affect delivery of the cancellation?
> 
> Also I'm uncomfortable with making two changes at once here, it feels as though the root cause hasn't been identified?

I've observed that after a certain period of time (Am currently measuring) the _rtcp_run loop will await at a specific point, after a lot of testing, I believe a version of the code I published worked, since rtcp survived overnight once, but I deleted that one. 

I've observed the above code does not work like I thought it would and the bug still appears, so I'll close this request. 
If you need full context or more info, forum here:
https://github.com/aiortc/aiortc/issues/792

PR #1115, opened 2024-06-02
Change 32 bit int longs to 64 bit qlongs to allow RTP to run for more…

R0NAM1:
… then two hours.

I've observed in my issues creating and closing camera webrtc streams for my NVR software that any stream that is over two hours long breaks. Over the course of about a month (After figuring out how to properly close the pc object- a whole other adventure) I figured out that the issue is a 32bit integer overflow in the octet_count of __bytes__ in RtcpSenderInfo.
The error presented was `__bytes exception: 'L' format requires 0 <= number <= 4294967295`.

Changing this to 64bit integers (Longs -> Qlongs) allows for 24 hours streams and possibly beyond instead of two hour streams. If more then 24 hours are wanted then people can do what I will: Run `pc.stop()` at midnight and then 30 seconds later refresh the webpage, all client side. Keeps it so no streams stay open forever.

If you think anything else is needed or required please let me know.
(Discussion thread https://github.com/aiortc/aiortc/issues/792)

### DIFF
--- a/src/aiortc/rtp.py
+++ b/src/aiortc/rtp.py
@@ -374,7 +374,7 @@ class RtcpSenderInfo:
 
     def __bytes__(self) -> bytes:
         return pack(
-            "!QLLL",
+            "!QQQQ",
             self.ntp_timestamp,
             self.rtp_timestamp,
             self.packet_count,


R0NAM1:
I should clarify the stream itself dosen't actually break, the RTCP part of it does, and this breaking means when I run pc.close() after it breaks the task never exits properly, so it causes runaway resources on my server.

jlaine:
This looks like an incorrect fix, RFC 3550 clearly states these are 32-bit, so the encoding is not the issue.

https://datatracker.ietf.org/doc/html/rfc3550#section-6.4.1

I think this needs to be clamped or wrapped around where we *create* the sender report:

https://github.com/aiortc/aiortc/blob/22699ea879f93b6d6dd1af4a200d37b9ff560870/src/aiortc/rtcrtpsender.py#L419-L420

I'd appreciate pointers to where his behaviour is specified in the RFC.

R0NAM1:
```
 sender's octet count: 32 bits
      The total number of payload octets (i.e., not including header or
      padding) transmitted in RTP data packets by the sender since
      starting transmission up until the time this SR packet was
      generated.  The count SHOULD be reset if the sender changes its
      SSRC identifier.  This field can be used to estimate the average
      payload data rate.
```
Ah that is annoying, then since the actual int type cannot be changed then we need to do what I tried originally, a quick fix that just keeps the rtcp_loop running. 

The main issue is that after two hours I can't stop the RTCPeer resources server side because the rtcp_loop freezes since bytes() errors out, perhaps just wrapping a try except block around bytes would still allow the loop to continue running?

From
```
    async def _send_rtcp(self, packets: List[AnyRtcpPacket]) -> None:
            payload = b""
            for packet in packets:
                self.__log_debug("> %s", packet)
                payload += bytes(packet)
                await self.transport._send_rtp(payload)
```
to
```
    async def _send_rtcp(self, packets: List[AnyRtcpPacket]) -> None:
        try:
            payload = b""
            for packet in packets:
                self.__log_debug("> %s", packet)
                payload += bytes(packet)
                await self.transport._send_rtp(payload)
        except ConnectionError:
            pass
```

or we can actually manipulate the __octet_count manually, which is the overflowing integer. Could we cap it's growth to it's integer limit so that way it can grow to it's max and then stay? I actually like this idea the best, let me know what you think however.

R0NAM1:
That could break other things though limiting it globally, another option is enforcing that limit only in _run_rtcp.

R0NAM1:
Just saw https://github.com/aiortc/aiortc/pull/1121/commits/7297cf6d6b63297518a9707dcead373b5a699268
Why do you believe the best option would be to wrap these integers around instead of capping the values? Is this also part of the RFC, so some standard?

PR #1117, opened 2024-06-10
Set correct codec for ogg output

andywer:
Use `libopus` if the output format as `ogg`. That's a pretty important case as WebRTC audio streams are usually opus streams.


andywer:
Ran into this issue when I was trying to record a WebRTC audio track for debugging.

jlaine:
This looks like a good catch, but it does need a unit test!

PR #1118, opened 2024-06-13
Fix reversed mid

hilookas:
If no break statement in the changed file, the last unmatched transceiver will get. That will cause mid order different from addtrack() order in the multiple video track situation which is counterintuitive. This commit fixes it by breaking the loop in advance.

### DIFF
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -843,6 +843,7 @@ class RTCPeerConnection(AsyncIOEventEmitter):
                 for t in self.__transceivers:
                     if t.kind == media.kind and t.mid in [None, media.rtp.muxId]:
                         transceiver = t
+                        break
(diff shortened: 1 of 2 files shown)


jlaine:
Could you please write a unit test for this?

jlaine:
@hilookas Do you plan to complete this PR with the requested unit test? Your fix may well be correct, but I'd like a test which captures the bug you are fixing.

hilookas:
> @hilookas Do you plan to complete this PR with the requested unit test? Your fix may well be correct, but I'd like a test which captures the bug you are fixing.

I added a new unit test. Please check it out.

shenwpo:
Will there be a new wheel? We have to replace `addTrack` with `addTransceiver` before PR merge.

jlaine:
Thanks for this. I have slightly refactored the test and will be merging the PR once CI passes.

PR #1123, opened 2024-06-21
Fix: DTLS handshake failed (fingerprintt mismatch)

defnull:
DTLS handshake fingerprint check should support more than just SHA-256.

BigBlueButton for example (based on Mediasoup / Freeswitch) adds SHA-384 fingerprints to the sdp offer. Connection works fine after applying this patch.

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -439,7 +439,7 @@ class RTCDtlsTransport(AsyncIOEventEmitter):
             if f.value.upper() == fingerprint:
                 break
         else:
-            self.__log_debug("x DTLS handshake failed (fingerprintt mismatch)")
+            self.__log_debug("x DTLS handshake failed (fingerprint mismatch)")
             self._set_state(State.FAILED)
             return
 
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -431,17 +431,15 @@ class RTCDtlsTransport(AsyncIOEventEmitter):
 
         # check remote fingerprint
         x509 = self._ssl.get_peer_certificate()
-        remote_fingerprint = certificate_digest(x509)
-        fingerprint_is_valid = False
         for f in remoteParameters.fingerprints:
-            if (
-                f.algorithm.lower() == "sha-256"
-                and f.value.lower() == remote_fingerprint.lower()
-            ):
-                fingerprint_is_valid = True
+            try:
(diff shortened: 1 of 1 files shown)


src/aiortc/rtcdtlstransport.py
```diff
@@ -431,17 +431,15 @@ async def start(self, remoteParameters: RTCDtlsParameters) -> None:
...
+            try:
+                fingerprint = x509.digest(f.algorithm).decode("ascii").upper()
+            except ValueError:
+                continue
+            if f.value.upper() == fingerprint:
                 break
-        if not fingerprint_is_valid:
-            self.__log_debug("x DTLS handshake failed (fingerprint mismatch)")
+        else:
+            self.__log_debug("x DTLS handshake failed (fingerprintt mismatch)")
```
fippo, line 442:
```suggestion
            self.__log_debug("x DTLS handshake failed (fingerprint mismatch)")
```

src/aiortc/rtcdtlstransport.py
```diff
@@ -431,17 +431,15 @@ async def start(self, remoteParameters: RTCDtlsParameters) -> None:
...
-        remote_fingerprint = certificate_digest(x509)
-        fingerprint_is_valid = False
         for f in remoteParameters.fingerprints:
-            if (
-                f.algorithm.lower() == "sha-256"
-                and f.value.lower() == remote_fingerprint.lower()
-            ):
-                fingerprint_is_valid = True
+            try:
+                fingerprint = x509.digest(f.algorithm).decode("ascii").upper()
```
fippo, line 436: probably good to check the algorithm against an allow list containing sha-256, sha-384 and sha-512 first.
Otherwise this would bring back sha-1 which nobody wants in 2024 ;-)

fippo (approved):

src/aiortc/rtcdtlstransport.py
defnull: I'm not sure if `SHA-1` is an issue in this context (it's just the certificate fingerprint) but a configurable whitelist would not hurt. But that would be a separate pull request and a bigger change?

jlaine:
I agree with @fippo, I'd prefer only using known-good hashes. Also I don't think the test should be "any valid fingerprint is OK", what if we have a mix of good / bad fingerprints?

From a high level I think our test should be : 
- Let supported algorithms be sha-256, sha-384, sha-512
- There must be at least one fingerprint presented by the remote party using a supported algorithm
- All supported fingerprints presented by the remote party must match

jlaine:
I've submitted an alternative implementation in #1216, what do you think @fippo?

PR #1152, opened 2024-09-07
Choose RTP sequence numbers in the lower half of 2**16

fippo:
to avoid the wraparound issue described by
  https://chromiumdash.appspot.com/commit/13b327b05fa3788b4daa9c3463e13282824cb320


fippo:
lint failures seem unrelated (but are automatically fixable by ruff)?

jlaine:
I've rebased your PR on `main` which fixes the linter issue. I've also reworked the code to add a helper which:

- Features an explicit comment on why we are doing this.
- Uses a modulo to avoid spurious code coverage errors.

PR #1157, opened 2024-09-17
Remux example

namoshizun:
Hii 👋

Here I added a simple example to illustrate how to remux a live media source (e.g., from RTSP) to use MPEGTS container, so that aiortc can stream it without transcoding.

In the example, I ran the remuxing in a child process executing plain ffmpeg command. I feel like it isn't the most ideal solution as it does assume the user has an existing ffmpeg. I tried [pyav's remuxing example](https://pyav.org/docs/develop/cookbook/basics.html#remuxing), but couldn't figure out how to properly create an output MPEGTS container that writes to some bytes buffer which then can be read and  parsed to generate pyav Packets...

### DIFF
--- a/examples/remux/README.rst
+++ b/examples/remux/README.rst
@@ -1,0 +1,56 @@
+Remux server 
+=============
+
+This example illustrates how to remux a video stream that is originally not packed in MPEGTS container so that it can be streamed without transcoding. An example usecase is playing the video content from an RTSP source that sends raw H265 streams.
+
+
+Running
+-------
+
+First install the required packages:
+
+.. code-block:: console
+
+    $ pip install aiohttp aiortc
+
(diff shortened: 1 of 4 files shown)


FreeNachman:
@namoshizun nice work, just curious, why we need the "extra" formatting/muxing into MPEGTS? can't we just de-mux packets from, say, RTSP source?

namoshizun:
> @namoshizun nice work, just curious, why we need the "extra" formatting/muxing into MPEGTS? can't we just de-mux packets from, say, RTSP source?

Yeah... actually later I found that it is possible to directly remux an RTSP stream without wrapping it in a MPEGTS container. But that is not always true for all RTSP streams. It works for some IPC cameras I have but not all of them.

Issue #580, opened 2021-10-28
non-strictly-monotonic PTS when recording


glebmxyz:
Hello! I've been working with aiortc a lot lately. Thank you for your project, @jlaine!

## The Problem
When I try to record video with server example i get the following output:
```
======== Running on http://0.0.0.0:8080 ========
(Press CTRL+C to quit)
INFO:pc:PeerConnection(5638e909-9064-42e6-b498-0b8b7f6ae66e) Created for 127.0.0.1
/home/glebmxyz/work/aiortc/examples/server/venv/lib/python3.9/site-packages/aiortc/rtcdtlstransport.py:211: CryptographyDeprecationWarning: This version of cryptography contains a temporary pyOpenSSL fallback path. Upgrade pyOpenSSL now.
  _openssl_assert(lib.SSL_CTX_use_certificate(ctx, self._cert._x509) == 1)  # type: ignore
INFO:pc:PeerConnection(5638e909-9064-42e6-b498-0b8b7f6ae66e) Track video received
/home/glebmxyz/work/aiortc/examples/server/venv/lib/python3.9/site-packages/aiortc/rtcdtlstransport.py:186: CryptographyDeprecationWarning: This version of cryptography contains a temporary pyOpenSSL fallback path. Upgrade pyOpenSSL now.
  value=certificate_digest(self._cert._x509),  # type: ignore
INFO:aioice.ice:Connection(0) Check CandidatePair(('192.168.88.254', 38006) -> ('192.168.88.254', 42846)) State.FROZEN -> State.WAITING
INFO:aioice.ice:Connection(0) Check CandidatePair(('172.17.0.1', 33384) -> ('192.168.88.254', 42846)) State.FROZEN -> State.WAITING
INFO:aioice.ice:Connection(0) Check CandidatePair(('172.19.0.1', 60979) -> ('192.168.88.254', 42846)) State.FROZEN -> State.WAITING
INFO:pc:PeerConnection(5638e909-9064-42e6-b498-0b8b7f6ae66e) Connection state is connecting
INFO:aioice.ice:Connection(0) Check CandidatePair(('192.168.88.254', 38006) -> ('192.168.88.254', 42846)) State.WAITING -> State.IN_PROGRESS
INFO:aioice.ice:Connection(0) Check CandidatePair(('192.168.88.254', 38006) -> ('192.168.88.254', 42846)) State.IN_PROGRESS -> State.SUCCEEDED
[...]

Issue #520, opened 2021-04-02
 janus example --record to udp stream has a timestamp doesn't increase issue


zhiweny1122:
Hi jlaine:
the udp mpegts stream from Janus example has a timestamp issue when I use the aiortc client to record to a janus server.
I pasted the log here.

root@wangzhiwen-Vostro-3668:/home/wangzhiwen/aiortc/examples/janus# python3.6 janus.py https://www.streamocean.com:8089/janus --record-to udp://172.16.0.31:1234
id: 8301182471340616, display: aa
Track audio received
Track video received
Exchanging media
[libx264 @ 0x557909691600] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2 AVX FMA3 BMI2 AVX2
[libx264 @ 0x557909691600] profile High, level 3.0, 4:2:0, 8-bit
[libx264 @ 0x557909691600] non-strictly-monotonic PTS
[libx264 @ 0x557909691600] non-strictly-monotonic PTS
[libx264 @ 0x557909691600] non-strictly-monotonic PTS
[libx264 @ 0x557909691600] non-strictly-monotonic PTS
[mpegts @ 0x5579094f03c0] Application provided invalid, non monotonically increasing dts to muxer in stream 1: 360000 >= 360000

zhiweny1122:
The janus public  website also has this issue.
the mpegts mux will stop if the timestamp is not correct. 
Can you take a look at it ?

zhiweny1122:
It looks like the pyav x264 encode create B frame and the pts has a non-strictly-monotonic PTS issue.
Can you tell me how to diable the B frame setting in codec ？

zhiweny1122:
Although I update the pts, the stream will stop in 10 minutes.

PR #1170, opened 2024-10-07
MediaRecorder: workaround for error "non-strictly-monotonic PTS error"

devluz:
Setting the time_base of the stream to the same of the incoming video tracks fixes errors some receive when recording incoming video tracks. 

Original problem:
The problem can be reproduced by sending a video stream with more than 30 FPS and attempt to record it via MediaRecorder. 
Typical error messages start with "non-strictly-monotonic PTS error" or " Application provided invalid, non monotonically increasing dts to muxer in stream ..." related to issues #580 and probably #520 

I suspect the root cause is that AV might default to a time base of 1/30 (if rate is set to 30). If the stream is received at 60FPS then two frames will share a timestamp once converted into the new time base. Many WebRTC based senders also do not enforce a fixed framerate so even 30FPS can randomly have two frames too close together and causing this issue. 

I tested this with both png and libx264 with no problems discovered so far.

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -13,7 +13,7 @@ from av.frame import Frame
 from av.packet import Packet
 from av.video.stream import VideoStream
 
-from ..mediastreams import AUDIO_PTIME, MediaStreamError, MediaStreamTrack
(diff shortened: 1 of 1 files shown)


sfarina:
This fixes an issue I have with consuming a janus stream.

Ioann44 (approved):

PR #1197, opened 2024-12-03
Add support for Python 3.13

gpauloski:
Adds support for Python 3.13 and drops support for EOL Python 3.8.

Feel free to edit or close if someone else is already working on this. I know this was unsolicited, but I was testing aiortc in Python 3.13 anyways and figured I'd share the changes.

There were two substantive changes:
1. Increase the upper limit for PyAV to support v13. This did cause an assertion in one test case to fail (`test_encoder_large`), but as far as I can tell there is no reason that the length of the payloads in that test must be one (nothing in the git blame indicated that either).
2. In Python 3.13, asyncio Unix sockets are removed when the server is closed so the explicit unlink needs to be skipped.

### DIFF
--- a/src/aiortc/contrib/signaling.py
+++ b/src/aiortc/contrib/signaling.py
@@ -167,7 +167,10 @@ class UnixSocketSignaling:
         if self._server is not None:
             self._server.close()
             self._server = None
-            os.unlink(self._path)
+            # In Python 3.13, asyncio Unix sockets are removed when the server is
+            # closed. On previous version we need to remove the socket ourselves.
+            if sys.version_info < (3, 13):
+                os.unlink(self._path)
 
     async def receive(self):
         await self._connect(False)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.9
       - name: Install OS packages
         run: |
           sudo apt-get update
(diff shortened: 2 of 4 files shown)


hpenedones:
This would be really useful for us -- we are implementing a new feature on the Inductiva API using aiortc, and we are bumping into this Python 3.13 compatibility problem. Is there any update?

hpenedones:
Who would be the natural reviewers for this change? @jlaine @uriyyo @rprata

bgregoinductiva:
> This would be really useful for us -- we are implementing a new feature on the Inductiva API using aiortc, and we are bumping into this Python 3.13 compatibility problem. Is there any update?

+1

jlaine:
I'm to trying to get CI back on its feet and will loop back to this PR, this is definitely something we want.

src/aiortc/contrib/signaling.py
```diff
@@ -167,7 +167,8 @@ async def close(self):
         if self._server is not None:
             self._server.close()
             self._server = None
-            os.unlink(self._path)
+            if sys.version_info < (3, 13):
+                os.unlink(self._path)
```
jlaine, line 171: Why do we need this? At the very least this requires a comment.

tests/test_h264.py
```diff
@@ -128,7 +128,7 @@ def test_encoder_large(self):
         # delta frame
         frame = self.create_video_frame(width=1280, height=720, pts=3000)
         payloads, timestamp = encoder.encode(frame)
-        self.assertEqual(len(payloads), 1)
+        self.assertGreaterEqual(len(payloads), 1)
```
jlaine, line 131: Is this actually related to Python 3.13 support? If it's not, please move it to a separate PR.

src/aiortc/contrib/signaling.py
jlaine: Sorry I saw your PR comment now and have integrated into the code.

tests/test_h264.py
jlaine: I'm yanking this change, but feel free to reintroduce it in a separate PR.

Issue #1137, opened 2024-08-01
Stopping a sender.track, removing sender from pc then stopping transceivers doesn't generate an track.end event


Nniol:
Hi,
Using the example server.py code as a base I have updated the solution so that when you click start it just opens a data channel and at this point only data is transmitted. I can stop and start this no problem. I can also choose to add in a microphone audio track and stream this audio to the server, all fine and I still have the data channel open to get responses about the data. I now choose to stop the audio track which I do via this code:
So I stop the track and remove the sender, I then stop the transceiver.
```
function stopRecording() {
    // close local audio / video
    var index = 1;
    pc.getSenders().forEach((sender) => {
        if (sender.track) {
            console.log ("Track stopped",index,sender.track);
            sender.track.stop();
            pc.removeTrack(sender);
            console.log ("Track stopped and removed",index,sender.track);    
        }
    });
    // close transceivers
    index = 1;
    if (pc.getTransceivers().length>0) {
        pc.getTransceivers().forEach((transceiver) => {
            console.log ("Transceiver stopping",index,transceiver.mid);
            transceiver.stop();
            console.log ("Transceiver stopped",index++,transceiver.mid)
        });
    }
    document.getElementById('use-audio').checked = false;
    negotiate();
    document.getElementById('start_recording').removeAttribute("disabled")
    document.getElementById('stop_recording').setAttribute("disabled","disabled")
}
```

This when I renegotiate I send this sdp, which as you can see has only one BUNDLE:  _a=group:BUNDLE 0_.
The sdp contains two mid sections: _a=mid:0_ and _a=mid:1_. I am assuming the mid:1 section is telling the server that the track is over and the transceivers are shutdown (_a=inactive_)
```
v=0
o=- 5738720534989457837 4 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0
a=extmap-allow-mixed
a=msid-semantic: WMS
m=application 57364 UDP/DTLS/SCTP webrtc-datachannel
c=IN IP4 192.168.153.1
a=candidate:1762407111 1 udp 2122260223 192.168.153.1 57364 typ host generation 0 network-id 3
a=candidate:2103215598 1 udp 2122194687 172.20.48.1 57365 typ host generation 0 network-id 4
a=candidate:1749320126 1 udp 2122129151 192.168.59.1 57366 typ host generation 0 network-id 5
[...]

Issue #1129, opened 2024-07-04
Using    sender.track.stop(); doesn't actually stop the recording or trigger the         @track.on("ended") code


Nniol:
I am using the example server code and want to keep the data connection open after I have finished recording. 
So I added a new button which appears after start is clicked upon, the function of this button is to call stop_recording() as shown below.

```
function stop_recording() {
    document.getElementById('stop_recording').style.display = 'none';
    document.getElementById('stop_session').style.display = 'inline-block';
    // close local audio / video
    pc.getSenders().forEach((sender) => {
        console.log ("sender",sender)
        sender.track.stop();
        console.log ("sender",sender)
    });
}

```

However when these are sent nothing happens in the server. The recording stops and no new data is sent, but the following code is not triggered

```
        @track.on("ended")
        async def on_ended():
            log_info("Track %s ended", track.kind)
            await recorder.stop()
```

When I click then click on Stop and trigger the stop() function then the recording stops, but so do the data connections which means I can't send anything back
```
function stop() {
    document.getElementById('stop_session').style.display = 'none';

    // close data channel
    if (dc) {
        dc.close();
    }
    // close transceivers
    if (pc.getTransceivers) {
        pc.getTransceivers().forEach((transceiver) => {
            if (transceiver.stop) {
                transceiver.stop();
            }
        });
    }

    // close peer connection
    setTimeout(() => {
        pc.close();
    }, 500);
}
```

Nniol:
I am assuming it is this bit of code in mediastreams.py that is the issue

```
class MediaStreamTrack(AsyncIOEventEmitter, metaclass=ABCMeta):
   
    ...
    ...
   
   
    def stop(self) -> None:
        if not self.__ended:
            self.__ended = True
            self.emit("ended")

            # no more events will be emitted, so remove all event listeners
            # to facilitate garbage collection.
            self.remove_all_listeners()
```

Nniol:
Ok, looking at an older issue I also now stop the transceivers and I am to stop recording, this removes the recording button from my web client, but still doesn't trigger the ended event in the server

```
function stop_recording() {
    // document.getElementById('stop_recording').style.display = 'none';
    // document.getElementById('stop_session').style.display = 'inline-block';
    // close local audio / video
    pc.getSenders().forEach((sender) => {
        console.log ("sender",sender)
        sender.track.stop();
        console.log ("sender",sender)
    });
    // close transceivers
    if (pc.getTransceivers) {
[...]

PR #1204, opened 2025-01-03
Stop transceiver when remote media direction is inactive

tyohan:
I have issue when trying to connect AioRTC client with Golang Pion based SFU that I made and I struggled to detect if remote track is stopped sending a media stream because even the remote track is stop and the remote peer which is Pion based SFU is negotiating by send an SDP with inactive direction in media SDP line, the AioRTC peer is doing nothing about it and never trigger track ended event.

I saw similar issues like [this](https://github.com/aiortc/aiortc/issues/1137), and [this](https://github.com/aiortc/aiortc/issues/1129) in AioRTC but no solutions so far.

I follow the implementation of Pion [RTCPeerConnection.SetRemoteDescription()](https://github.com/pion/webrtc/blob/1ee02999eb6ce157c291fc2dbf0d43bf5c700c71/peerconnection.go#L1090) which will check if a remote media has inactive direction and stop the related local transceiver which will trigger the track ended event.

### DIFF
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -475,7 +475,7 @@ class RTCPeerConnection(AsyncIOEventEmitter):
             direction=direction, kind=kind, sender_track=track
         )
 
-    async def close(self):
+    async def close(self) -> None:
         """
         Terminate the ICE agent, ending ICE processing and streams.
         """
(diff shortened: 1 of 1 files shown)


fippo:
"inactive" means "I am not sending or receiving media currently", stopping the track is not the right thing to do in this case
stopping a transceiver means "I am never going to use this again"

jlaine:
> "inactive" means "I am not sending or receiving media currently", stopping the track is not the right thing to do in this case stopping a transceiver means "I am never going to use this again"

I fully agree!

PR #1216, opened 2025-01-29
Add support for sha-384 and sha-512 certificate fingerprints

jlaine:
We now support SHA-256, SHA-384 and SHA-512 certificate fingerprints:

- We generate fingerprints for all supported algorithms.
- When checking fingerprints, we verify there is at least one supported algorithm, and that all supported fingerprints match.

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -37,6 +37,14 @@ CERTIFICATE_T = TypeVar("CERTIFICATE_T", bound="RTCCertificate")
 
 logger = logging.getLogger(__name__)
 
+# Mapping of supported `RTCDtlsFingerprint` algorithms to the
+# corresponding argument for `x509.digest`.
+X509_DIGEST_ALGORITHMS = {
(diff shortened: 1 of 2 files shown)


defnull:
LGTM, thanks :)

jlaine:
Thanks to you for getting the ball rolling on this and sorry for the slow réponse time

fippo:
> We generate fingerprints for all supported algorithms.

That is rather uncommon and I am not even sure what browsers will do (or should do) - from the look of libWebRTC i'd say "discards everything but the last" which should work out.

Be conservative what gets sent out (e.g. sha-256) but accept all of them?

fippo:
I checked and it kinda works. However, Chrome / libWebRTC only stores the last fingerprint which can be seen if you do
```
const pc = new RTCPeerConnection();
await pc.setRemoteDescription({type: 'offer', sdp: 'sdp from aiortc'})
console.log(pc.remoteDescription.sdp)
```
which will not show the sha-256 and sha-384 hashes in the SDP.

PR #1229, opened 2025-01-31
implement bundlePolicy

### LINKED ISSUE #555: bundlePolicy support
Web browser support bundlePolicy set to "max-bundle"  now,

can we support it?

fippo:
fixes #555

Notably this also changes the default behavior to share the same ice-ufrag/ice-pwd for all transceivers and the sctp transport, aligning with browsers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "aioice>=0.9.0,<1.0.0",
+    "aioice>=0.10.1,<1.0.0",
     "av>=14.0.0,<15.0.0",
     "cffi>=1.0.0",
     "cryptography>=44.0.0",
--- a/src/aiortc/rtcconfiguration.py
+++ b/src/aiortc/rtcconfiguration.py
@@ -1,3 +1,4 @@
+import enum
 from dataclasses import dataclass
 from typing import Optional, Union
 
@@ -22,6 +23,35 @@ class RTCIceServer:
     credentialType: str = "password"
 
 
+class RTCBundlePolicy(enum.Enum):
+    """
+    The :class:`RTCBundlePolicy` affects which media tracks are negotiated if
+    the remote endpoint is not bundle-aware, and what ICE candidates are
+    gathered.
+
+    See https://w3c.github.io/webrtc-pc/#rtcbundlepolicy-enum
+    """
+
+    BALANCED = "balanced"
+    """
+    Gather ICE candidates for each media type in use (audio, video, and data).
+    If the remote endpoint is not bundle-aware, negotiate only one audio and
+    video track on separate transports.
+    """
+
+    MAX_COMPAT = "max-compat"
+    """
+    Gather ICE candidates for each track. If the remote endpoint is not
+    bundle-aware, negotiate all media tracks on separate transports.
+    """
+
+    MAX_BUNDLE = "max-bundle"
+    """
+    Gather ICE candidates for only one track. If the remote endpoint is not
+    bundle-aware, negotiate only one media track.
+    """
+
+
 @dataclass
 class RTCConfiguration:
     """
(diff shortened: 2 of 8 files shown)


jlaine:
Heh, I'm not sure where this is headed: we're adding a configuration option, but not supporting it? :)

fippo:
Baby steps. Actual support isn't that far out I think

src/aiortc/rtcpeerconnection.py
```diff
@@ -1056,6 +1061,18 @@ def __createDtlsTransport(self) -> RTCDtlsTransport:
         # create ICE transport
         iceGatherer = RTCIceGatherer(iceServers=self.__configuration.iceServers)
         iceGatherer.on("statechange", self.__updateIceGatheringState)
+        if (
+            self.__configuration.bundlePolicy == RTCBundlePolicy.BALANCED
+            and len(self.__transceivers) > 0
+        ):
+            # ICE credentials are shared in "balanced" mode.
+            # TODO: fix aioice to take parameters instead
```
fippo, line 1069: This is a hack obviously and we would need an update to aioice to construct a gatherer with local username and password

Note that "balanced" does not mean "shares a single gatherer", this is observable in Chrome which with the default bundle gathers different local candidates.

max-bundle actually shares a single gatherer

fippo:
This turned out to be interesting because of how bundlePolicy i defined [here](https://w3c.github.io/webrtc-pc/#dom-rtcbundlepolicy):
* balanced: generate candidates for the first (audio, video, data) m= line respectively.
* max-compat: generate candidates for each m-line
* max-bundle: generate candidates only for the first m-line
In all cases the ice-ufrag and ice-pwd are the same. I believe what aiortc currently does would be permissible for max-compat but hey, lets just get over it ;-)

And ofc there is [**the contradiction**](https://www.rfc-editor.org/rfc/rfc8843.html#name-contradiction-regarding-bun)

src/aiortc/rtcpeerconnection.py
fippo: @jlaine shall I do a PR for aioice? I am a bit torn since this is not in the ORTC spec and should not be necessary to share but this is how all browsers behave and the more aligned one is with them...

pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "aioice>=0.9.0,<1.0.0",
+    "aioice>=0.10.1,<1.0.0",
```
fippo, line 28: necessary since we depend on 0.10.x to pass username and password to the gatherer.

fippo:
Took a bit longer but ... PTAL!

src/aiortc/rtcconfiguration.py
```diff
@@ -31,3 +42,5 @@ class RTCConfiguration:
 
     iceServers: Optional[list[RTCIceServer]] = None
     "A list of :class:`RTCIceServer` objects to configure STUN / TURN servers."
+
+    bundlePolicy: Optional[RTCBundlePolicy] = RTCBundlePolicy.BALANCED
```
jlaine, line 46: Is this actually optional? As in: if I set bundlePolicy to None does it work?

src/aiortc/rtcrtptransceiver.py
```diff
@@ -130,6 +130,13 @@ async def stop(self) -> None:
         await self.__sender.stop()
         self.__stopped = True
 
+    @property
+    def transport(self) -> RTCDtlsTransport:
```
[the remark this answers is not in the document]
jlaine: I don't see a `transport` attribute here?

https://w3c.github.io/webrtc-pc/#rtcrtptransceiver-interface

jlaine:
This is fantastic @fippo thanks so much! I added the documentation strings so that they show up here:

https://aiortc.readthedocs.io/en/latest/api.html#aiortc.RTCBundlePolicy

I'm going to cut a 1.12.0 release ASAP

fippo:
heh. Took a bit longer and scope grew from "lets just use the same ufrag" but ... :-)

src/aiortc/rtcrtptransceiver.py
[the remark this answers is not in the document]
fippo: ah I was in the wrong file / object.

PR #1246, opened 2025-02-01
github actions: depend on lint and test for building wheels

fippo:
building wheels does not make much sense until lint and tests pass


jlaine:
TBH I'm not sure we need this: the wheels will not be *published* unless all tests pass. I quite like having an exhaustive view of what's passing / failing

fippo:
Mostly being mindful in using githubs resources ;-)

jlaine:
Let's go, +1 for saving resources.

PR #1254, opened 2025-02-02
Update opus codec to use PyAV instead of libopus

dodobyte:
Didn't remove the libopus dependency, if you're happy with the PR you can remove it separately.


jlaine:
> Didn't remove the libopus dependency, if you're happy with the PR you can remove it separately.

Haha, so kind of you, but I'd rather have a self contained PR.

Anyway I think we're not suite there, the test suite is failing. You're probably going to have to rework packetisation and / or account for buffering due to the resampler.

PR #1255, opened 2025-02-20
Add `types-pyOpenSSL` to lint deps

sobolevn:
Looks like stubs for this packages were not installed. It might be an error in the future.

See https://github.com/python/mypy/pull/18372


jlaine:
PyOpenSSL features built-in type annotations since 24.2.0, so there is no need for `types-PyOpenSSL` anymore. I should know, I added them in:

https://github.com/pyca/pyopenssl/pull/1308

PR #1257, opened 2025-02-21
Fix RTX packet decoding error

edhyah:
Original problem:
When packets are dropped or lost, RTX packets are sent in WebRTC. In aiortc, these packets, when received by RTCRtpReceiver, errored at runtime citing "ValueError: No decoder found for MIME type `video/rtx`". This happens because after RTX packets are unwrapped, the codec isn't updated to the apt codec, which should be used to decode the unwrapped RTX packet.

This fix updates the codec that will be used to decode the unwrapped RTX packet to the apt codec.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,36 @@
+FROM pytorch/pytorch:2.4.0-cuda11.8-cudnn9-runtime
+
+# Install build dependencies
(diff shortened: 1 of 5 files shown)


fippo:
same as #1260 which has less stuff that is unrelated?

jlaine:
Closing in favour of #1260

Issue #1258, opened 2025-02-21
'video/rtx' MIMEType retransmission incorrectly handled as a codec needing a decoder


ahundt:
Expected behavior:
When parts of a video stream are retransmitted, that info should be re-incorporated into the main stream.

Behavior encountered:
Retransmitted data is interpreted as an additional separate stream, so decoding fails with the error below.

```
Exception in thread video-decoder:
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/threading.py", line 1041, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "/usr/local/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/threading.py", line 992, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/athundt/source/gemini-audio-video-chat/.venv/lib/python3.13/site-packages/aiortc/rtcrtpreceiver.py", line 65, in decoder_worker
    decoder = get_decoder(codec)
  File "/Users/athundt/source/gemini-audio-video-chat/.venv/lib/python3.13/site-packages/aiortc/codecs/__init__.py", line 155, in get_decoder
    raise ValueError(f"No decoder found for MIME type `{mimeType}`")
ValueError: No decoder found for MIME type `video/rtx`
2025-02-21 17:41:27,130 - INFO - Video emit cancelled.
2025-02-21 17:41:27,133 - WARNING - RTCRtpsender(video) Traceback (most recent call last):
  File "/Users/athundt/source/gemini-audio-video-chat/.venv/lib/python3.13/site-packages/aiortc/rtcrtpsender.py", line 359, in _run_rtp
    enc_frame = await self._next_encoded_frame(codec)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/athundt/source/gemini-audio-video-chat/.venv/lib/python3.13/site-packages/aiortc/rtcrtpsender.py", line 311, in _next_encoded_frame
    payloads, timestamp = self.__encoder.pack(data)
                          ~~~~~~~~~~~~~~~~~~~^^^^^^
[...]

PR #1260, opened 2025-02-25
Fix: No decoder found for MIME type video/rtx retransmission

kontorol:
https://github.com/aiortc/aiortc/issues/1258#issuecomment-2679917663

Issue: ValueError: No decoder found for MIME type video/rtx

Description

A ValueError is being raised in the decoder_worker function due to the absence of a decoder for the MIME type video/rtx. This issue occurs when handling RTP packets in the RTCRtpReceiver class.

Description

A ValueError is being raised in the decoder_worker function due to the absence of a decoder for the MIME type video/rtx. This issue occurs when handling RTP packets in the RTCRtpReceiver class. The error specifically happens when a track is added to MediaRelay.

Suggested Fix

This happens because after RTX packets are unwrapped, the codec isn't updated to the apt codec, which should be used to decode the unwrapped RTX packet.
The line 'codec = self.__codecs[apt]' should be added at line 509 to update the codec.

```python
async def _handle_rtp_packet(self, packet: RtpPacket, arrival_time_ms: int) -> None:
    ...
    # unwrap retransmission packet
    if is_rtx(codec):
        original_ssrc = self.__rtx_ssrc.get(packet.ssrc)
        if original_ssrc is None:
            self.__log_debug("x RTX packet from unknown SSRC %d", packet.ssrc)
            return

        apt = codec.parameters.get("apt")
        if (
            len(packet.payload) < 2
            or not isinstance(apt, int)
            or apt not in self.__codecs
        ):
            return

        packet = unwrap_rtx(packet, payload_type=apt, ssrc=original_ssrc)
        codec = self.__codecs[apt]
    ...
    
   ```

### DIFF
--- a/src/aiortc/rtcrtpreceiver.py
+++ b/src/aiortc/rtcrtpreceiver.py
@@ -506,6 +506,7 @@ class RTCRtpReceiver:
                 return
 
             packet = unwrap_rtx(packet, payload_type=apt, ssrc=original_ssrc)
+            codec = self.__codecs[apt]
 
         # send NACKs for any missing any packets
         if self.__nack_generator is not None and self.__nack_generator.add(packet):


freddyaboulton (approved):

jlaine:
This looks good, I'm wondering if we could have a follow-up PR with a test exercising this condition

Issue #688, opened 2022-04-06
Ungraceful shutdown in webcam example


alegonz:
Hi and thanks for this great library.

As stated in the title, I noticed that I get this `RuntimeError: Event loop is closed` error when I hit Ctrl+C while streaming with a webcam using the `webcam.py` example.

```shell
~$ python webcam.py
# Hit Ctrl+C

(...snip...)

^CConnection state is closed
Exception in thread media-player:
Traceback (most recent call last):
  File "/Users/foo.user/.pyenv/versions/3.9.10/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/Users/foo.user/.pyenv/versions/3.9.10/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/foo.user/0_work/code/aiortc/.venv/lib/python3.9/site-packages/aiortc/contrib/media.py", line 153, in player_worker
    asyncio.run_coroutine_threadsafe(video_track._queue.put(frame), loop)
  File "/Users/foo.user/.pyenv/versions/3.9.10/lib/python3.9/asyncio/tasks.py", line 934, in run_coroutine_threadsafe
    loop.call_soon_threadsafe(callback)
  File "/Users/foo.user/.pyenv/versions/3.9.10/lib/python3.9/asyncio/base_events.py", line 791, in call_soon_threadsafe
    self._check_closed()
  File "/Users/foo.user/.pyenv/versions/3.9.10/lib/python3.9/asyncio/base_events.py", line 510, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
/Users/foo.user/.pyenv/versions/3.9.10/lib/python3.9/threading.py:975: RuntimeWarning: coroutine 'Queue.put' was never awaited
  self._invoke_excepthook(self)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

I tried stopping the relay/track before closing event loop with an appropriate SIGINT handler, but that didn't solve the issue.

However, I noticed that, if I remove the `MediaRelay` and instead pass `webcam.video` directly to `pc.addTrack`, the example still works and shuts down gracefully with Ctrl+C.

What could be happening here? Why does one need `MediaRelay` to use a webcam?

I'm using `aiortc==1.3.1` on Python 3.9.10 on MacOS.

Thank you!

RyantheKing:
[...]

PR #1267, opened 2025-03-02
fix runtime error on ungraceful shutdown

RyantheKing:
Fix an issue where MediaPlayer won't exit properly on a ctrl+c shutdown.
Issue described here: [https://github.com/aiortc/aiortc/issues/688](https://github.com/aiortc/aiortc/issues/688)
(Occurs in webcam example)

### DIFF
--- a/src/aiortc/contrib/media.py
+++ b/src/aiortc/contrib/media.py
@@ -138,7 +138,10 @@ def player_worker_decode(
                 audio_samples += frame.samples
 
                 frame_time = frame.time
-                asyncio.run_coroutine_threadsafe(audio_track._queue.put(frame), loop)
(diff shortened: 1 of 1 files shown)


SergeantSerk:
Tried with `aiortc @ git+https://github.com/aiortc/aiortc.git@4814ca30a83a00eac244b0ef4252487db5153baf`. No longer encountering the described issue.

I was previously coming across it due to another issue `[video4linux2,v4l2 @ 0x11ef1240] Some buffers are still owned by the caller on close` and thought it would fix it but that seems to be due to another problem. This PR fixes the linked issue for me.

jlaine:
I provided an actual fix for the underlying issue in #1328

PR #1285, opened 2025-03-30
Mark `RTCIceServer.urls` as being a list of strings

jlaine:
This is an alternative proposal to #1270.


jlaine:
Let's keep it simple, the initial proposal was fine

PR #862, opened 2023-04-10
Allow adding candidates with only sdpMLineIndex

eerimoq:
Didn't implement any tests yet. Will do if the fix itself will be accepted.

PR #1300, opened 2025-05-04
Fix for addIceCandidate not working with only sdpMLineIndex

kjjohnsen:
This fix addresses a problem where some ice candidates trickle with only sdpMLineIndex set.  The documentation from addIceCandidate suggests that this works, but there was no code handling that case.  

This is based on https://github.com/aiortc/aiortc/pull/862 , but includes a unit test to confirm functionality

### DIFF
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -432,8 +432,8 @@ class RTCRtpSender:
                         sender_info=RtcpSenderInfo(
                             ntp_timestamp=self.__ntp_timestamp,
                             rtp_timestamp=self.__rtp_timestamp,
-                            packet_count=self.__packet_count,
-                            octet_count=self.__octet_count,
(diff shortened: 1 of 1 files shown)


src/aiortc/rtcpeerconnection.py
```diff
@@ -406,14 +406,21 @@ async def addIceCandidate(self, candidate: RTCIceCandidate) -> None:
...
+                candidate.sdpMid == transceiver.mid
+                or candidate.sdpMLineIndex == transceiver._get_mline_index()
+            ) and not transceiver._bundled:
                 iceTransport = transceiver._transport.transport
                 await iceTransport.addRemoteCandidate(candidate)
                 return
 
         if (
             self.__sctp
             and candidate.sdpMid == self.__sctp.mid
```
jlaine, line 419: Unless I'm mistaken this looks wrong, one of the conditions is duplicated?

jlaine:
Ouch sorry I messed up my push...

PR #1302, opened 2025-05-07
fix: Fix the octet_count overflow issue by modifying the _run_rtcp function of the RTCRtpSender class

Rocloud1999:
When the RTCRtpSender operates continuously for approximately 2 hours or longer, the octet_count in RtcpSenderInfo may overflow. This can cause the RTCP component to cease transmission, thereby disrupting the audio-video synchronization mechanism and resulting in severe desynchronization during playback at the receiver's end.

The error displayed was `__bytes exception: 'L' format requires 0 <= number <= 4294967295.`

In Chrome's WebRTC C++ implementation, octet_count is implicitly converted to the uint32_t type, effectively performing a high-bit truncation. In Python, the same logic can be achieved using `& 0xFFFFFFFF` , thereby preventing overflow issues.

### DIFF
--- a/src/aiortc/rtcrtpsender.py
+++ b/src/aiortc/rtcrtpsender.py
@@ -432,8 +432,8 @@ class RTCRtpSender:
                         sender_info=RtcpSenderInfo(
                             ntp_timestamp=self.__ntp_timestamp,
                             rtp_timestamp=self.__rtp_timestamp,
-                            packet_count=self.__packet_count,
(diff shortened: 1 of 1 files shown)


jlaine:
This is a good catch thank you!

I was wondering whether it would be better to directly store the truncated value in `self.__octet_count` and `self.__packet_count`, but these counters are also used in the `getStats()` API which does not need to perform the truncation, so I think your approach is the right one.

PR #1304, opened 2025-05-11
Fix for addIceCandidate not working with only sdpMLineIndex

jlaine:


jlaine:
This superseeds #862 and #1300

jlaine:
@kjjohnsen I've taken a different approach for the unit tests: we already have some detailed tests which do both RTP and SCTP using trickle. We create a variant of these tests which uses the m-line index instead of the mid.

PR #1306, opened 2025-05-11
avoid having a transceiver._transport

fippo:
instead accesss it via the receivers transport. This is somewhat arbitrary since senders and receivers always exist but similar to how a transceivers receiver track is always set so you can get the "transceiver kind" from it.

Also remove transceiver.transport accessor added in
  https://github.com/aiortc/aiortc/pull/1229#discussion_r1938353163


fippo:
Semantic changes, no actual logic changes

PR #1307, opened 2025-05-11
Split up RTCDtlsTransport.start()

fippo:
to make the individual steps more clear

(quite interesting that the DTLS transport has no getRemoteParameters() and hence does not store them)

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -406,38 +406,10 @@ class RTCDtlsTransport(AsyncIOEventEmitter):
             fingerprints=self.__local_certificate.getFingerprints()
         )
 
-    async def start(self, remoteParameters: RTCDtlsParameters) -> None:
+    async def _do_handshake(self) -> None:
         """
-        Start DTLS transport negotiation with the parameters of the remote
-        DTLS transport.
-
-        :param remoteParameters: An :class:`RTCDtlsParameters`.
+        Attempt to complete the DTLS handshake.
         """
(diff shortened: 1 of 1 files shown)


jlaine:
I'm about to land a *large* patch which tightens our type definitions, so this is unfortunately going to need rebasing when #1309 lands :)

fippo:
Rebased without merge conflicts!

src/aiortc/rtcdtlstransport.py
```diff
@@ -515,6 +493,53 @@ async def start(self, remoteParameters: RTCDtlsParameters) -> None:
...
+        # Initialise SSL.
+        self._ssl = SSL.Connection(
+            self.__local_certificate._create_ssl_context(
+                srtp_profiles=self._srtp_profiles
+            )
+        )
+        if self._role == "server":
+            self._ssl.set_accept_state()
+        else:
+            self._ssl.set_connect_state()
```
jlaine, line 525: How come this is not part of `do_handshake`?

jlaine:
While I don't have a strong reason *not* to merge this, I'm kind of lacking an incentive to do so, do you have any follow up changes in mind?

fippo:
Prototyping DTLS-in-STUN described [here](https://issues.webrtc.org/issues/367395350#comment1).
This requires changes to the order of starting ICE and DTLS (currently: DTLS after ICE; with that: the other way round) and I believe (but haven't finished the PoC) this will be much easier to understand with DTLS steps split up (OTOH the configuration would also need to go to a separate method)

PR #1313, opened 2025-05-21
Limit the number of threads for VP8 encoding

jlaine:
When we switched to using PyAV for VP8 encoding we dropped the logic which defined the number of threads to use for encoding. This results in using as many threads as there are CPUs:

https://github.com/FFmpeg/FFmpeg/blob/b6f84cd72acd84ba725922bd6a8967b416b2230a/libavcodec/libvpxenc.c#L1014

Restore the previous behaviour.

### DIFF
--- a/src/aiortc/codecs/vpx.py
+++ b/src/aiortc/codecs/vpx.py
@@ -1,3 +1,4 @@
+import multiprocessing
 import random
 from struct import pack, unpack_from
 from typing import Optional, Type, TypeVar, cast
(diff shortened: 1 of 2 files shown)


fippo (approved):
Looks familiar. Made me wonder if we recreate the encoder when the input resolution changes which was something that resulted (due to different number of threads without reinit) a CVE against Chrome a year or two ago (but only WebCodecs, libWebRTC did it correctly)

PR #1317, opened 2025-05-22
Add support for G.722 audio

jlaine:
The G.722 codec is "special" because the clock rate is 8kHz even though the sampling rate is 16kHz.


vonsky104:
@jlaine Hello, is there some plan to include that support in some 'soonish' release? If yes, is there some date of such release maybe? Thanks in advance.

PR #1318, opened 2025-05-23
add support for addIceCandidate(null) and update remote description with candidates added

### LINKED ISSUE #1299: End of candidates type error
I am connecting to JS WebRTC, after gathering it gives end of candidates with empty candidate but still has sdpMid and sdpMLineIndex, so I can still make the candidate class with it and the rest is empty. The addIceCandidate() checks whether there is sdpMid and sdpMLineIndex:

https://github.com/aiortc/aiortc/blob/9664b669bcee50dc41e341656b2f8ab19ce69623/src/aiortc/rtcpeerconnection.py#L405-L406

next it will pass the candidate to addRemoteCandidate(), but in there, it checks if the candidate is None as the end of candidates, but the parameter is always a class object in addIceCandidate() so it will never know if its end of candidates:

https://github.com/aiortc/aiortc/blob/9664b669bcee50dc41e341656b2f8ab19ce69623/src/aiortc/rtcicetransport.py#L282-L291

the addIceCandidate() requires the RTCIceCandidate as parameter, even if I pass a None type, it will be cause an error because it checks the sdpMid and sdpMLineIndex. aiortc gatherer doesn't seem to make end of candidates, but Isn't it better to either check for end of candidates in addIceCandidate() too or checks for sdpMid and sdpMLineIndex on the addRemoteCandidate() instead of addIceCandidate()?

Sorry if I made a mistake, I'm a beginner

fippo:
fixes #1299


fippo:
While reading code, [setting `self.__iceGatherer._remote_candidates_end`](https://github.com/aiortc/aiortc/blob/main/src/aiortc/rtcicetransport.py#L302) seems semantically wrong, the gatherer is not aware of remote candidates?

src/aiortc/rtcpeerconnection.py
```diff
@@ -425,6 +433,17 @@ async def addIceCandidate(self, candidate: RTCIceCandidate) -> None:
             iceTransport = self.__sctp.transport.transport
             await iceTransport.addRemoteCandidate(candidate)
 
+        # Update the remote description.
+        media = self.__remoteDescription().media
+        for sdp_m_line_index in range(0, len(media)):
+            if not candidate:
```
jlaine, line 439: if `candidate is None` for consistency?

jlaine:
> While reading code, [setting `self.__iceGatherer._remote_candidates_end`](https://github.com/aiortc/aiortc/blob/main/src/aiortc/rtcicetransport.py#L302) seems semantically wrong, the gatherer is not aware of remote candidates?

It doesn't know the actual candidates, but we do need somewhere to store the fact we received an end-of-candidates. As ICE transports constructed from the same ICE gatherer share the same connection, this seemed like a natural place to store this information. Alternatively we could modify `aioice.Connection` to expose an accessor for this information, since it already stores it:

https://github.com/aiortc/aioice/blob/29976f5415714df13787b933012f8a966a4859bf/src/aioice/ice.py#L491

PR #1320, opened 2025-05-23
Support creating an offer without media or data channels

fippo:
while not very useful this is how browsers behave after
  https://github.com/w3c/webrtc-pc/issues/1745

(ah, those were the days...)


fippo:
better safe then sorry, everything worked out thankfully ;-)

jlaine:
What actually gets negotiated in this case, should we be inspecting the SDP? Sorry for the slow drip reviews..

fippo:
just a session part. What is confusing about this case in general is that due to lack of ICE candidates this will never connect.

PR #1363, opened 2025-09-06
Add vp9

hmaarrfk:
Thanks for the cool software.

I'm trying to experiment with different target bitrates and codecs.

I still can't get VP9 or H264 to work on any browser of mine, but even https://mozilla.github.io/webrtc-landing/pc_test.html show that forcing H264 doesn't seem to work...

I generally think that you'll have many people trying to tweak parameters like this, and the ability to adjust things without digging so far down the chain will be helpful. Let me know what you think I can split the PR however you wish.


hmaarrfk:
i might want to add VP9 support, but I need to find the right encoding parameters.

We also need to consider the fact that the options for the differe "h264" encoders are 100% different, with basically the only one being shared is the "level" which might not even be shared well between them.

PR #1379, opened 2025-10-13
add support for RTCConfiguration.alwaysNegotiateDataChannels

fippo:
which always negotiates datachannels and puts them first in the SDP. Spec:
  https://w3c.github.io/webrtc-extensions/#always-negotiating-datachannels


jlaine:
Can we maybe have some comments in the code about the order in which the sections are added?

PR #1392, opened 2025-11-13
additional cipher suite for gstreamer webrtc server

FabienDanieau:
Using aiortc as client for the gstreamer webrtc component leads to the following error:

```
0:00:19.629139112  3949   0x7f1c003510 ERROR         dtlsconnection gstdtlsconnection.c:994:handle_error:<dtlsconnection0> Fatal SSL error
0:00:19.629283927  3949   0x7f1c003510 ERROR         dtlsconnection gstdtlsconnection.c:977:ssl_err_cb:<dtlsconnection0> ssl error: 60F1F9047F000000:error:0A0000C1:SSL routines:tls_post_process_client_hello:no shared cipher:../ssl/statem/statem_srvr.c:2333:

0:00:19.629456576  3949   0x7f1c003510 ERROR                dtlsdec gstdtlsdec.c:503:process_buffer:<dtlsdec0> Error processing buffer: Fatal SSL error
```

adding ECDHE-RSA-AES128-GCM-SHA256 to the cipher list fixes the issue.

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -201,7 +201,7 @@ class RTCCertificate:
         ctx.use_certificate(self._cert)
         ctx.use_privatekey(self._key)
         ctx.set_cipher_list(
-            b"ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA"
+            b"ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES128-GCM-SHA256"
         )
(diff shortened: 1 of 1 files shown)


fippo:
https://datatracker.ietf.org/doc/html/rfc8827#section-6.5 makes TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 mandatory to implement so I wonder why gstreamer does not support that?

FabienDanieau:
[Here's their answer](https://discourse.gstreamer.org/t/webrtc-cipher-suite-compatibility/5542/2). It's indeed not implemented on their side

fippo:
The SRTP cipher can't be the problem. This looks like
  https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/dtls/gstdtlsagent.c?ref_type=heads#L201
isn't playing nicely, possibly adding [DEFAULT](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/rtc_base/openssl_stream_adapter.cc;l=1162?q=SSL_CTX_set_cipher_list%20webrtc) as done in libWebRTC will do the trick?

FabienDanieau:
DEFAULT works! Do you want me to change the PR?

PR #1396, opened 2025-11-19
sdp: handle missing a=msid-semantic: line

### LINKED ISSUE #1395: Should not depend on "a=msid-semantic: WMS" deprecated attribute
aiortc heavily relies on the presence of `a=msid-semantic: WMS ..." session attribute to later identify remote tracks via their corresponding media level `a=msid` attributes.

This is done here: https://github.com/aiortc/aiortc/blob/156db1488af18d07a742dd2db68465f366765537/src/aiortc/sdp.py#L566

```py
    def webrtc_track_id(self, media: MediaDescription) -> Optional[str]:
        assert media in self.media
        if media.msid is not None and " " in media.msid:
            bits = media.msid.split()
            for group in self.msid_semantic:
                if group.semantic == "WMS" and (
                    bits[0] in group.items or "*" in group.items
                ):
                    return bits[1]
        return None
```

However, `a=msid-semantic: WMS` attribute was removed in the final RFC 8830, see: https://datatracker.ietf.org/doc/html/rfc8830#section-appendix.a

AFAIS it existed until **draft-ietf-mmusic-msid-08** (this is 2015, **10 years ago**): https://datatracker.ietf.org/doc/html/draft-ietf-mmusic-msid-08#section-3

fippo:
which is a thing of the past despite still being sent by browsers in various different forms. If not present, assume wildcard semantics.

Fixes #1395

### DIFF
--- a/tests/test_sdp.py
+++ b/tests/test_sdp.py
@@ -592,6 +592,32 @@ a=setup:active
             ),
         )
 
+    def test_no_msid_semantic(self) -> None:
+        d = SessionDescription.parse(
+            lf2crlf(
+                """v=0
+o=- 863426017819471768 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+a=group:BUNDLE 0
+m=audio 16628 UDP/TLS/RTP/SAVPF 8
+a=rtpmap:8 PCMA/8000
+a=fingerprint:sha-256 35:5A:BC:8E:CD:F8:CD:EB:36:00:BB:C4:C3:33:54:B5:9B:70:3C:E9:C4:33:8F:39:3C:4B:5B:5C:AD:88:12:2B
+a=setup:active
+a=rtcp-mux
+a=rtcp:16628 IN IP4 1.2.3.4
+a=ice-ufrag:75EDuLTEOkEUd3cu
+a=ice-pwd:5dvb9SbfooWc49814CupdeTS
+a=candidate:0560693492 1 udp 659136 1.2.3.4 16628 typ host generation 0
+a=end-of-candidates
+a=msid:stream deprecatedtrack
+"""
+            )
+        )
+
+        self.assertEqual(d.media[0].msid, "stream deprecatedtrack")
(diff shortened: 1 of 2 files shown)


tests/test_sdp.py
```diff
@@ -592,6 +592,32 @@ def test_audio_freeswitch(self) -> None:
...
+a=candidate:0560693492 1 udp 659136 1.2.3.4 16628 typ host generation 0
+a=end-of-candidates
+a=msid:stream deprecatedtrack
+"""
+            )
+        )
+
+        self.assertEqual(d.media[0].msid, "stream deprecatedtrack")
+        self.assertEqual(d.webrtc_track_id(d.media[0]), "deprecatedtrack")
+
```
jlaine: How is this going to be serialised back?

fippo:
```
v=0
o=- 863426017819471768 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0
m=audio 16628 UDP/TLS/RTP/SAVPF 8
a=msid:stream deprecatedtrack
a=rtcp:16628 IN IP4 1.2.3.4
a=rtcp-mux
a=rtpmap:8 PCMA/8000
a=candidate:0560693492 1 udp 659136 1.2.3.4 16628 typ host
a=end-of-candidates
a=ice-ufrag:75EDuLTEOkEUd3cu
a=ice-pwd:5dvb9SbfooWc49814CupdeTS
a=fingerprint:sha-256 35:5A:BC:8E:CD:F8:CD:EB:36...
a=setup:active
```
which is what the standard says. Now browsers don't do what the standard say but that is a different problem :cry:

jlaine: Well maybe actually test it? 😆

fippo: oh you mean how the O/A behavior is? E.g. remote offer without msid-semantic but with msid (an sdp string for the test) and the question is whether we respond with a msid-semantic of our own?

fippo: Finally got around to doing it!

PR #1390, opened 2025-10-17
Add VP9 video codec support

himmelroman:
First off - I LOVE aiortc! I've been using it for years with great success in production. Thank you to all the maintainers and contributors for this amazing library! 🙏

## Summary

This PR adds full VP9 video codec support to aiortc, following RFC 9628 (RTP Payload Format for VP9 Video). The implementation is an exact copy of Pion WebRTC's VP9 implementation in Go, ensuring compatibility and correctness.

## Features

- ✅ Full RFC 9628 compliance for parsing (all modes supported)
- ✅ Both flexible mode (F=1) and non-flexible mode (F=0) encoding
- ✅ VP9 header parser for bitstream analysis
- ✅ Scalability Structure (V=1) support on keyframes
- ✅ Browser-tested with Chrome (both modes working)
- ✅ Configurable via `flexible_mode` parameter (defaults to True for maximum browser compatibility)

### Core Components

**Vp9PayloadDescriptor:**
- Complete RFC 9628 payload descriptor with full serialization/parsing
- Supports 7-bit and 15-bit Picture IDs
- Layer indices (TID, SID, U, D flags)
- Reference indices (P_DIFF) for flexible mode
- TL0PICIDX for non-flexible mode
- Scalability Structure parsing (V=1)

**Vp9Encoder:**
- Uses PyAV's libvpx-vp9 with WebRTC-optimized settings
- Configurable flexible/non-flexible mode
- Picture ID and TL0PICIDX management
- Bitrate control and multi-threaded encoding

**Vp9Decoder:**
- libvpx-vp9 decoder integration
- RTP depacketization
- Jitter buffer compatibility

**VP9 Header Parser:**
- `_parse_vp9_header()`: Bitstream parser for keyframe/inter-frame detection
- Extracts frame dimensions from keyframes
- EXACT copy of Pion's header.go logic

**Packetization Methods:**
- `_packetize_flexible()`: EXACT copy of Pion's payloadFlexible()
- `_packetize_non_flexible()`: EXACT copy of Pion's payloadNonFlexible()

## Testing

- ✅ 25 new VP9 unit tests (13 descriptor + 12 encoder/decoder)
- ✅ 1 integration test in test_rtcpeerconnection.py
- ✅ All 519 aiortc tests passing
- ✅ Browser compatibility verified with Chrome (both flexible and non-flexible modes)

## Files Modified

- `src/aiortc/codecs/vpx.py` (+~700 lines) - VP9 implementation
[...]

PR #1401, opened 2025-12-06
Fix payload type collision in codec negotiation

himmelroman:
## Summary

Fixes payload type (PT) collision issues during SDP offer/answer negotiation, implementing libwebrtc's collision resolution algorithm.

### Problem

When negotiating codecs, aiortc could assign the same payload type to multiple codecs. This happens when:
1. Remote SDP has duplicate PTs (e.g., RTX codec and main codec both using PT 103)
2. BUNDLE groups share a PT namespace across audio/video m-lines

Browsers reject SDP with duplicate payload types, causing connection failures.

### Solution

- **PT collision resolution**: When a preferred PT is already in use, search for an unused PT in the dynamic range [96-127], descending from 127 (matching libwebrtc behavior)
- **BUNDLE namespace tracking**: Pass a shared `used_payload_types` set across media negotiation when BUNDLE is active (RFC 8834)
- **RTX apt updates**: When a base codec's PT is reassigned, update the corresponding RTX codec's `apt` parameter to match

### Changes

**`rtcpeerconnection.py`**:
- Add `used_payload_types` parameter to `find_common_codecs()` for session-wide PT tracking
- Implement `find_unused_id()` - searches [127→96] for available PT
- Implement `find_and_set_id_used()` - assigns PT with collision handling
- Track BUNDLE groups and share PT namespace across media lines
- Update RTX `apt` parameter when base codec PT changes

**`test_rtcpeerconnection.py`**:
- `test_find_common_codecs_pt_collision_rtx_and_codec`: RTX/codec collision scenario
- `test_find_common_codecs_pt_collision_descending`: Multiple collisions use descending search
- `test_find_common_codecs_bundle_collision`: Cross-media PT collision in BUNDLE

### Related

This fix was originally part of #1390 (VP9 support) and has been split into a separate PR per maintainer request.

### DIFF
--- a/src/aiortc/rtcpeerconnection.py
+++ b/src/aiortc/rtcpeerconnection.py
@@ -82,30 +82,71 @@ def filter_preferred_codecs(
 def find_common_codecs(
     local_codecs: list[RTCRtpCodecParameters],
     remote_codecs: list[RTCRtpCodecParameters],
+    used_payload_types: Optional[set[int]] = None,
 ) -> list[RTCRtpCodecParameters]:
+    """
+    Find common codecs between local and remote, negotiating payload types.
+
+    Implements libwebrtc's PT collision resolution algorithm. When used_payload_types
+    is provided, maintains session-wide PT namespace for BUNDLE (RFC 8834).
+    """
     common = []
-    common_base: dict[int, RTCRtpCodecParameters] = {}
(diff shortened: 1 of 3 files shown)


himmelroman:
@jlaine Are there other maintainers I can ping for this?

jlaine:
There is quite a bit of churn in the commits here, would it be possible to break this into descriptive commits for each change?

PR #1402, opened 2025-12-08
feat: Add iceTransportPolicy support to RTCConfiguration

### LINKED ISSUE #1397: Feature request: Support for setting the ICE transport policy (relay-only mode)
Can the library be updated to allow the transport policy to be set when gathering ice candidates?

I noticed that aioice supports setting the transport_policy via https://github.com/aiortc/aioice/pull/68, but aiortc does not currently have a way to set this parameter (https://github.com/aiortc/aiortc/blob/main/src/aiortc/rtcicetransport.py#L196)

Our use case is that our deployment environment requires the WebRTC connection to flow through TURN relay servers rather than attempting direct peer-to-peer connections (similar to the browser's https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#icetransportpolicy)

Thank you!

rkrdo:
Implements `iceTransportPolicy` for `RTCConfiguration` to control ICE candidate gathering, including a new `RTCIceTransportPolicy` enum and integration with `aioice.Connection`.

Closes #1397

### DIFF
--- a/tests/test_rtcicetransport.py
+++ b/tests/test_rtcicetransport.py
@@ -1,11 +1,12 @@
 import asyncio
-from typing import Optional
+from typing import Any, Optional
+from unittest.mock import AsyncMock, patch
 
 import aioice.ice
 import aioice.stun
 from aioice import ConnectionClosed
 from aiortc.exceptions import InvalidStateError
-from aiortc.rtcconfiguration import RTCIceServer
+from aiortc.rtcconfiguration import RTCIceServer, RTCIceTransportPolicy
 from aiortc.rtcicetransport import (
     RTCIceCandidate,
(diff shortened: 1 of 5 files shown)


tests/test_rtcicetransport.py
```diff
@@ -249,7 +260,6 @@ async def test_gather(self) -> None:
         self.assertEqual(gatherer.state, "completed")
         self.assertTrue(len(gatherer.getLocalCandidates()) > 0)
 
-        # close
```
jlaine, line 252: Why this change?

rkrdo: that wasn't intentional, i reverted the deletion

PR #1417, opened 2026-04-12
Reset decoder CodecContext on error and send PLI

tokoroten:
## Summary

- Reset VP8/H264 `CodecContext` after decode error to clear corrupted reference frames
- Add `decode_errors` counter to decoders for error signaling
- `decoder_worker` detects decode failures and signals via `threading.Event`
- `RTCRtpReceiver` sends RTCP PLI when decoder error is detected

## Problem

When a video decoder encounters an `FFmpegError` (e.g., from corrupted data due to packet loss), three things go wrong:

1. **CodecContext is not reset** — the decoder retains corrupted reference frames and produces garbage output for all subsequent frames until a keyframe arrives
2. **No PLI is sent** — the receiver never requests a keyframe from the sender, so recovery depends entirely on natural keyframe intervals
3. **Errors are silently swallowed** — `decode()` catches the exception, logs a warning, and returns `[]` with no feedback to the receiver

Together, these cause **permanent video corruption** after packet loss in practice.

### Decoder reset (`vpx.py`, `h264.py`)

```python
# Before: error swallowed, codec reused in corrupted state
except av.FFmpegError as e:
    logger.warning("...skipping package: " + str(e))
    return []

# After: codec recreated, error counter incremented
except av.FFmpegError as e:
    logger.warning("...resetting: " + str(e))
    self.codec = CodecContext.create("libvpx", "r")  # fresh decoder
    self.decode_errors += 1
    return []
```

### PLI signaling (`rtcrtpreceiver.py`)

`decoder_worker` checks `decoder.decode_errors` after each decode call. On error, it sets a `threading.Event` that `RTCRtpReceiver._handle_rtp_packet()` checks on the next packet arrival, triggering an RTCP PLI.

## Test plan

- [x] `test_decode_error_resets_codec` — CodecContext is recreated after error (VP8/H264)
- [x] `test_multiple_errors_recreate_codec_each_time` — each error creates a fresh codec
- [x] `test_pli_event_set_on_decode_error` — decoder_worker signals PLI on error
- [x] `test_pli_event_not_set_on_no_error` — no false PLI on normal operation
- [x] `test_vp8/h264_decoder_tracks_errors` — decode_errors attribute exists and increments
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/aiortc/codecs/h264.py
+++ b/src/aiortc/codecs/h264.py
@@ -118,9 +118,13 @@ class H264Decoder(Decoder):
             self.decode_errors = 0
             return frames
         except av.FFmpegError as e:
-            logger.warning("H264Decoder() failed to decode, resetting: %s", e)
-            self.codec = av.CodecContext.create("h264", "r")
             self.decode_errors += 1
+            if self.decode_errors == 1:
+                # First error: reset codec and signal for PLI.
+                # Subsequent errors are expected (P-frames without reference)
+                # until a keyframe arrives — drop silently to avoid PLI storm.
+                logger.warning("H264Decoder() failed to decode, resetting: %s", e)
+                self.codec = av.CodecContext.create("h264", "r")
             return []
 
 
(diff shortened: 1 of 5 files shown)


jlaine:
I agree we need a trigger to send PLI. However what do we gain by resetting the codec context? Isn't this just guaranteed to trigger errors on every subsequent packet (and hence reset the codec context again) until the peer reacts to the PLI?

How do browsers handle this situation?

PR #1421, opened 2026-05-19
Add RSA ciphers for compatibility with RSA-based servers

helloworldsuper:
Description
Currently, aiortc limits DTLS cipher suites to ECDSA only. This causes handshake failures when interacting with media servers or endpoints that use RSA certificates, such as ZLMediaKit (especially via WHEP).

Changes
I have extended the set_cipher_list in RTCDtlsTransport to include ECDHE-RSA-AES128-GCM-SHA256 and ECDHE-RSA-AES256-GCM-SHA384. These are standard suites that fix the interoperability issue while maintaining high security.

Reference
Relates to ZLMediaKit compatibility issues (e.g.,(https://github.com/ZLMediaKit/ZLMediaKit/issues/4559)) .

### DIFF
--- a/src/aiortc/rtcdtlstransport.py
+++ b/src/aiortc/rtcdtlstransport.py
@@ -201,7 +201,7 @@ class RTCCertificate:
         ctx.use_certificate(self._cert)
         ctx.use_privatekey(self._key)
         ctx.set_cipher_list(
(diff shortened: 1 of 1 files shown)


jlaine:
I don't think we will be merging this, it explicitly goes against this PR:

https://github.com/aiortc/aiortc/pull/1011

@fippo thoughts?

fippo:
While https://www.rfc-editor.org/rfc/rfc8827.html#name-communications-security mandates support for TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 I don't see much harm in additional suites.

fippo (approved):

PR #1432, opened 2026-08-06
Use the DTLS role to pick the SCTP role

Rickaym:
`is_server` on `RTCSctpTransport` is taken from the ICE role today. The spec (RFC 8832) says it should follow the DTLS role: whoever is the DTLS client acts as the SCTP client and uses the even stream IDs.

Most of the time the two roles line up, so this rarely bites. But they can differ. We hit it on production voice calls where our side answered with `setup:active` (DTLS client) while being ICE controlled. The remote peer followed the spec, so both ends decided they were the server, nobody sent INIT, and the data channel just never opened. Audio kept working, which made it fun to debug.

The fix reads the DTLS role instead and flips the even/odd stream ID assignment to match. Two aiortc peers still end up on opposite roles, so nothing changes between them. The fix matters when talking to an implementation that follows the RFC strictly.

I had to swap the client/server pair in the test fixture (the "client" was built on the ICE-controlling side, which is the DTLS server) and update a few stream ID assertions to the new parity. Full suite passes.


fallenmi (approved):
The SCTP association role and DCEP stream-ID parity now follow the negotiated DTLS role, as required by RFC 8832, instead of being inferred from ICE. On exact head `64459234b61caba0190f8dbb958ca4ba8dd6a581`, I exercised all four combinations of ICE controlling/controlled and explicit DTLS client/server: the DTLS client initiated SCTP and selected even stream ID 0, while the DTLS server waited and selected odd stream ID 1 in every case. The same oracle passes on GitHub synthetic merge `8651ae0f8ff1b1fe79c374cc3179afa96853edc5`; exact base `8a2864630bf417d977a06dda8d0254b1c1501bfb` fails all four role/parity/initiator combinations.

The pristine exact-head SCTP suite passed 72/72, and the full suite passed 495/495. Ruff check/format, check-manifest, and mypy also pass, and the exact head and merge trees are identical. I found no blocking issue.

AI disclosure: I used OpenAI Codex to inspect the DTLS/SCTP startup and DCEP stream-allocation paths, compare them with RFC 8832 and RFC 8261, construct and run the exact base/head/merge role matrix, execute the focused/full test and static-check suites, and draft this review. I manually verified the role combinations, observed values, exact commits/tree, public policy and interactions, specifications, and conclusion.

PR #1433, opened 2026-08-10
Add bytesReceived to RTCInboundRtpStreamStats

cancan101:
# Add `bytesReceived` to `RTCInboundRtpStreamStats`

## What

`RTCRtpReceiver.getStats()` currently reports `packetsReceived`, `packetsLost`
and `jitter` for an inbound stream, but no byte counter — the
`RTCInboundRtpStreamStats` dataclass is an empty subclass and the construction
site in `rtcrtpreceiver.py` ends with a bare `# RTPInboundRtpStreamStats`
comment with nothing under it.

This adds `bytesReceived`:

* `RTCInboundRtpStreamStats.bytesReceived: int` in `stats.py`;
* `StreamStatistics.bytes_received`, accumulated in `StreamStatistics.add()`,
  which already receives the `RtpPacket`;
* the value wired through in `RTCRtpReceiver.getStats()`.

It also fixes the typo in the adjacent comment marker
(`RTPInboundRtpStreamStats` → `RTCInboundRtpStreamStats`), since the new field
is filed under it.

## Why

`bytesReceived` is the counter you need to answer "how much bandwidth is this
stream costing", and it is the natural counterpart to the already-implemented
`RTCOutboundRtpStreamStats.bytesSent`. Every browser reports it, and
webrtc-stats treats it as part of the baseline set — its Conformance section's
sample wording for referencing specifications (marked "EXAMPLE ONLY", so
illustrative rather than binding on aiortc) reads:

> An implementation MUST support generating statistics for the type
> `RTCInboundRtpStreamStats`, with members `packetsReceived`, `bytesReceived`,
> `packetsLost`, and `jitter`.
>
> — [webrtc-stats §2 Conformance](https://www.w3.org/TR/webrtc-stats/#conformance)

Without it, a receiver-side application has to reach into `RTCTransportStats`,
which is per-transport and therefore aggregates every bundled stream plus
RTCP/DTLS/STUN overhead.

## Semantics: what is counted

**Payload bytes only — no RTP header, no padding.** That is the spec's
definition, made explicit by the way `headerBytesReceived` is defined against
it:

> `bytesReceived` of type `unsigned long long`
> Total number of bytes received for this SSRC. This includes retransmissions.
> Calculated as defined in [RFC3550] section 6.4.1.
>
> `headerBytesReceived` of type `unsigned long long`
> Total number of RTP header and padding bytes received for this SSRC. This
> includes retransmissions. This does not include the size of transport layer
> headers such as IP or UDP. `headerBytesReceived` + `bytesReceived` equals the
> number of bytes received as payload over the transport.
>
> — [webrtc-stats §8.4 RTCInboundRtpStreamStats](https://www.w3.org/TR/webrtc-stats/#dom-rtcinboundrtpstreamstats-bytesreceived)

RFC 3550 §6.4.1 (the referenced definition, for the symmetric sender octet
count) says the same thing in the other direction: "The total number of payload
octets (i.e., not including header or padding) transmitted in RTP data packets".

This maps cleanly onto aiortc: `RtpPacket.parse()` already strips padding into
`padding_size` and leaves `payload` holding exactly the payload octets, so the
counter is `len(packet.payload)`. It is also symmetric with the existing sender
side, which accumulates `bytesSent` as `len(payload)` in
`RTCRtpSender._run_rtp()`.

Consequences worth stating explicitly:

* A **padding-only** packet increments `packetsReceived` by 1 and
  `bytesReceived` by 0. Covered by a test.
* **Header extensions and CSRCs** are not counted, as they are header bytes.

## Judgement calls a reviewer should check

1. **Placement on `RTCInboundRtpStreamStats`, not `RTCReceivedRtpStreamStats`.**
   The spec puts `bytesReceived` on `RTCInboundRtpStreamStats`;
   `RTCReceivedRtpStreamStats` carries only `packetsReceived` / `packetsLost` /
   `jitter`. That is also the practical choice here: the other subclass,
   `RTCRemoteInboundRtpStreamStats`, is built from RTCP Receiver Reports, which
   carry no byte count, so hoisting the field would force an invented value
   there.

2. **RTX is counted where aiortc already counts packets, i.e. under the RTX
   SSRC.** `StreamStatistics.add()` is called before `unwrap_rtx()`, keyed on
   the wire SSRC, so an RTX stream gets its own `StreamStatistics` — and the
   retransmitted payload bytes land there, not folded into the media stream's
   counter. The spec says RTX "streams do not show up as separate
   `RTCInboundRtpStreamStats` objects but affect the `packetsReceived`,
   `bytesReceived`, `retransmittedPacketsReceived` and
   `retransmittedBytesReceived` counters of the relevant
   `RTCInboundRtpStreamStats` objects"
   ([§5.1](https://www.w3.org/TR/webrtc-stats/#dfn-inbound-rtp)).
   I deliberately did **not** change that: `packetsReceived` already behaves
   this way, and re-attributing RTX to the media SSRC would mean reworking
   existing accounting (and the RTCP Receiver Report the same object feeds,
   which legitimately needs per-SSRC sequence/jitter state for the RTX stream).
   Byte counting is kept consistent with packet counting; happy to follow up
   separately if you would rather the stats layer diverged from the RTCP layer
   here.

3. **No `headerBytesReceived`.** Deliberately kept minimal. It would be easy to
   add (`packet` knows its own header size only indirectly, so it would need a
   little more plumbing than `len(payload)`), but it is a separate, optional
   member.

4. **No changelog entry.** `docs/changelog.rst` looks like it is written at
   release time; say the word and I will add one.

5. Pre-existing and untouched: every inbound stream shares the stats id
   `"inbound-rtp_" + str(id(self))`, so with multiple SSRCs on one receiver
   (e.g. media + RTX) the entries overwrite each other in the `RTCStatsReport`.
   That already affects `packetsReceived`; this PR does not make it worse, and
   fixing it changes stats ids, so it seemed out of scope.

## Tests

Three additions to `tests/test_rtcrtpreceiver.py`:

* `StreamStatisticsTest.test_bytes_received` — 10 packets × 160-byte payload
  accumulate to `bytes_received == 1600`, alongside `packets_received == 10`.
* `StreamStatisticsTest.test_bytes_received_padding_only` — feeds the existing
  `rtp_only_padding.bin` fixture (236 bytes on the wire, 224 of them padding,
  zero-length payload) and asserts `bytes_received == 0` while
  `packets_received == 1`, pinning the payload-only semantics above.
* `RTCRtpReceiverTest.test_rtp_and_rtcp` — asserts the counter reaches
  `getStats()`: after 10 `rtp.bin` packets the `inbound-rtp` entry reports
  `packetsReceived == 10` and `bytesReceived == 1600`.

The shared `create_rtp_packets()` helper gained an optional `payload` argument
(defaulting to `b""`, so existing callers are unaffected).

## Checks

Run locally on macOS / CPython 3.14, matching `.github/workflows/tests.yml`:

* `coverage run -m unittest discover` — 497 tests, OK.
* `ruff check .` — all checks passed.
* `ruff format --check --diff .` — 70 files already formatted.
* `check-manifest` — lists match.
* `mypy src tests` — no issues found in 59 source files.
* `make -C docs html SPHINXOPTS=-W` — build succeeded (the new attribute
  docstring renders under the existing `autoclass:: RTCInboundRtpStreamStats`).


fallenmi (changes requested):
The new `bytesReceived` counter is reported on the wrong inbound stream when RTX is negotiated. `_handle_rtp_packet()` updates `StreamStatistics` using the wire SSRC before `unwrap_rtx()` maps the packet back to its media SSRC. `getStats()` then gives every stream the same `inbound-rtp_<receiver-id>` id, so the later RTX entry replaces the media entry in `RTCStatsReport`.

I reproduced this deterministically on exact head `8dde9e12520431bfcdc9b6dd568d747dc4c7dceb`. With media SSRC `1234`, RTX SSRC `2345`, one six-byte media payload, and one eight-byte RTX payload, `getStats()` returns `(ssrc, packetsReceived, bytesReceived) = (2345, 1, 8)`. The inbound media statistic should remain attached to SSRC `1234` and include both received packets and their payload bytes. The W3C stats model is explicit that RTX streams do not appear as separate `RTCInboundRtpStreamStats` objects and instead affect the counters of the associated media object: https://www.w3.org/TR/webrtc-stats/#dfn-inbound-rtp

Please keep any RTCP-specific per-wire-SSRC sequence/jitter accounting you need, but build the public inbound-media counter after the RTX association so RTX contributes to the relevant media stat and cannot overwrite it. Please also add a media+RTX `getStats()` regression.

The same oracle passes on exact base `8a2864630bf417d977a06dda8d0254b1c1501bfb` with the new field absent. The exact synthetic merge `bb118d47951b4a3d9978c2228033687fc1d30c38` passes all 497 existing tests, plus `ruff`, `check-manifest`, and `mypy`; the submitted tests only exercise a single non-RTX SSRC and therefore miss this case.

AI disclosure: I used OpenAI Codex to inspect the receiver/RTX accounting and stats identity paths, check the W3C specification, construct and run the exact base/head oracle, execute the full synthetic-merge test and lint suites, and draft this review. I manually verified the code paths, observed values, exact commits, live state, and conclusion.

cancan101:
Good catch — thanks. You're right that statistics are recorded per wire SSRC before
`unwrap_rtx()` runs. The RTX packets weren't mutating the media stream's
`StreamStatistics` (each wire SSRC gets its own entry), but the effect you describe was
real: `getStats()` builds every entry with the same report id, so with RTX active the
RTX stream's entry overwrote the media stream's `inbound-rtp` entry in the report —
leaving it with the RTX SSRC and RTX-only counters, whose byte count also included the
2-byte OSN header.

Fixed in cc658c9:

- `getStats()` now skips RTX SSRCs, so the RTX stream no longer surfaces as a separate
  `inbound-rtp` entry (per webrtc-stats).
- Each recovered packet and its payload bytes (counted after `unwrap_rtx()` strips the
  OSN header) are credited to the original SSRC, so the media entry's
  `packetsReceived`/`bytesReceived` include retransmissions — consistent with
  `retransmittedPacketsReceived`/`retransmittedBytesReceived` being defined as subsets
  of those counters.
- The recovered counts are deliberately tracked apart from `StreamStatistics`: those
  objects also feed the RTCP receiver reports, where RFC 4588 treats the RTX stream as
  a distinct RTP stream with its own sequence space, and folding recoveries in there
  would distort the media stream's reported loss and jitter.
- Added `test_rtp_rtx_stats` covering your exact scenario: media packets 0 and 2 on
  SSRC 1234, packet 1 recovered via RTX on SSRC 2345 → the report contains a single
  `inbound-rtp` entry attached to SSRC 1234 with `packetsReceived == 3` and the
  recovered payload bytes included.

Note this overwrite predates this PR — `packetsReceived`, `packetsLost`, and `jitter`
had the same collision, and the shared `"inbound-rtp_" + id(self)` report id would
still collide for a receiver with multiple non-RTX streams. I kept the id format
unchanged here to avoid widening the PR; happy to fix that separately (e.g. including
the SSRC in the id) if the maintainers want.

PR #1440, opened 2026-09-19
sctp: bump max-message-size to 256k

fippo:
because things have progressed a bit since the mess described in
  https://lgrahl.de/articles/demystifying-webrtc-dc-size-limit.html
This matches the default Chrome and Firefox use these days. 64kb is still used as default remote MMS if none is announced.

Throw an error when attempting to send something larger than the MMS announced by the remote.

Note: the inbound MMS is not (yet) enforced


fippo:
@lgrahl time flies, doesn't it?

fippo:
since .send() can now throw I wonder if that needs a version bump.

lgrahl:
> since .send() can now throw I wonder if that needs a version bump.

Good question. What would have happened before if one were to send a DC message that is too large?

fippo:
apparently dcsctp on the other side will say "nay" if one sends larger than MMS, basically what is still missing after this.

lgrahl:
That means it would basically not have worked anyways, right? Then I don't think we need a (major) version bump.

fippo:
Different failure mode. Oversized send would have caused an async disconnect from the remote while now it throws.

Fairly certain nobody does this and it is the spec behavior.

lgrahl:
True, slightly different failure modes. But personally, I think an async failure in combination with the rarity (probably noone does this, like you said) is sufficient here to rationalise not having to bump the version.

PR #1441, opened 2026-09-19
sctp: reduce the MTU to 1189 bytes

fippo:
see https://issues.webrtc.org/issues/42222640
This is the size of the SCTP packet including the 12 byte common header, leaving 1160 bytes of user data per DATA chunk after the 16 byte chunk header and 4-byte padding.

See also
  https://github.com/pion/sctp/pull/476
and the funny tailscale story


lgrahl:
The calculation is still not exactly accurate since DTLS overhead can vary between versions and ciphers (in theory, at least). But it's a step up.

lgrahl (approved):

fippo:
and DTLS 1.3 reduces the overhead but since we are on OpenSSL we don't get that yet...

